PR #3, opened 2014-02-03
Add CMake buildfile.

berkus:
Very simple CMakeLists.txt to allow building librtaudio.a with cmake.


radarsat1:
@derkus can you check #77 please?

PR #7, opened 2014-04-11
Allow out-of-tree builds

Ace17:
Here's a very simple patch allowing to do this (out of tree builds):

svn checkout svn://svn.code.sf.net/p/rtaudio/code/trunk rtaudio-code
cd rtaudio-code
mkdir build
cd build
../configure
make
make -C tests

@top_srcdir@ and @srcdir@ are predefined variables. They get replaced by autotools (make doesn't see them).


garyscavone:
This is very old now (and has conflicts) and there didn't seem to be further interest from others, so I'm closing it.

PR #33, opened 2015-03-29
Fix ALSA underrun warning message

igor-mikushkin:
This fixes underrun warning text in error callback for ALSA API.
In some circumstances a text was not set.
That caused a usage of previous text.


garyscavone:
Manually merged

PR #38, opened 2015-04-14
Fix CMake compilation on Linux

ColinGilbert:


garyscavone:
Would this change create a problem for people that don't have the most up-to-date compiler?  Why is the change necessary?

ColinGilbert:
Honestly, I do not know. I believe that gcc 4.6 can use C++11 and from what I read on forums, it seems that almost no-one runs with a lesser version. Perhaps a conditional check for older compilers might help? All I know is that it wouldn't work on my system (gcc 4.8.3) without that flag.

radarsat1:
Why does this patch include a bunch of CMake-generated files?  Also, since RtAudio does not have a dependency on C++11, I don't recommend "fixing" anything by enable C++11 support.

radarsat1:
Sorry, I understand that it adds and then removes a bunch of CMake-generated files.  In the future, it would be nice to clean up the history before submitting a pull request.

ColinGilbert:
Sorry, it was one of first times doing a pull request

berkus:
I think this can be closed as invalid, it includes generated files and adds -std=c++11 to build flags which is not really required (although supported by rtaudio if necessary).

garyscavone:
Closing for reasons given above.

Issue #52, opened 2016-03-11
Split RtAudio.cpp into multiple files


joshdekock:
As of writing this, RtAudio.cpp has 10230 lines, and as far as I can tell, it's the only source file which makes up RtAudio itself.
I'm sure this makes it very difficult to navigate (it definitely did for me), so wouldn't splitting it into multiple files, say, by class make it nicer to work with?

garyscavone:
Personally, I prefer the single file.  It makes it much easier to search and edit similar code structures in the various classes.  And I prefer only needing to copy two files into a project when I want realtime audio support.

joshdekock:
It was just a thought, I guess if people prefer having it all in one file, then it's not my place to try change that. Although, it might be a good idea to have some sort of index: an order of functions, so that, people who don't know what functions are in the file (and where they are), can quickly see an overview.

pffang:
I think it's a good idea.
Splitting it into multiple files can make the codes more clear.

cjcliffe:
I'm a fan of the single monolithic source -- mostly for simple for project inclusion.  Documentation at http://www.music.mcgill.ca/~gary/rtaudio/ was sufficient to implement without much need to read the source files but might be nice to have in the GitHub wiki for contributions and updates.

If it was ever broken up I'd still prefer to have a build target or script to compose it back into a single file like some javascript library release builders do.

joshdekock:
> If it was ever broken up I'd still prefer to have a build target or script to compose it back into a single file like some javascript library release builders do.

This could be a viable compromise.

However, this brings me on to my next question, why isn't RtAudio distributed as a library? Surely this would remove the need for it to be in one source file, as people would link to the (static) library instead; they have to link to an object file anyways.

DouglasHeriot:
RtAudio being one giant file is one reason I’ve hesitated to use it – it makes it harder for me to come at it and understand the structure. Before I start using a library, I need to be comfortable maintaining & bugfixing it if needed.

I like how SQLite solves this – you can develop it split into as many files as you want, and there’s a makefile that amalgamates it into a single file for easy inclusion. https://www.sqlite.org/amalgamation.html

radarsat1:
> However, this brings me on to my next question, why isn't RtAudio distributed as a library?

The build does generate a library?

    $ make >/dev/null 2>&1 && find . -name \*.so
    ./.libs/librtaudio.so

joshdekock:
[...]

PR #56, opened 2016-03-28
RtAudio.cpp: split into multiple files for better organisation

joshdekock:
This fixes issue #52. It separates the files for better organisation, and adds a target to both the cmake and autotools build systems: `make unify` which will result in a `RtAudioUnified.cpp` file being created (which is pretty much exactly the same as the `RtAudio.cpp` file currently)


garyscavone:
I pretty much said a while back that I would not support splitting RtAudio.cpp.  As long as I control the master branch, I won't approve it.  Simply said, I prefer a single file for coding and I do not see enough substantiation for changing it.  Within a normal editor like Emacs, it should not be a problem to deal with a long file and it is much simpler to have a single file to implement (or check) common functionality across the various APIs.

PR #65, opened 2016-06-14
Includes saturation to avoid buffer overflow on integer conversion

guidefloripa:
Includes saturation to avoid buffer overflow on integer conversion

### DIFF
--- a/RtAudio.cpp
+++ b/RtAudio.cpp
@@ -9922,7 +9922,7 @@ void RtApi :: convertBuffer( char *outBuffer, char *inBuffer, ConvertInfo &info 
       Float32 *in = (Float32 *)inBuffer;
       for (unsigned int i=0; i<stream_.bufferSize; i++) {
         for (j=0; j<info.channels; j++) {
-          out[info.outOffset[j]] = (Int32) (in[info.inOffset[j]] * 2147483647.5 - 0.5);
+          out[info.outOffset[j]] = (Int32) std::max(-2147483648.0, std::min(2147483647.0, in[info.inOffset[j]] * 2147483647.5 - 0.5));
         }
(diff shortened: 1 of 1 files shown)


speps:
This should be done by the user of the library by clamping from -1.0 to 1.0. That conversion method should be as lean as possible in my opinion adding a min/max on each conversion is branching on each value...

guidefloripa:
You're right. This could be done by the user.

guidefloripa:
One more detail: when you have a float audio backend like on MacOS, it's possible to overflow the -1.0 to +1.0 range to get a "louder" and not distorted sound at output. However, in other 16-bit backends, the range -1.0 to +1.0 must be respected because overflow causes distortions.

As RtAudio is a multiplatform library, and allows the user to choose the sample type, I think the saturation in the buffer conversion keeps RtAudio more flexible and allows the user overflows the range when it's supported (in float backends) and saturated when the overflow causes distortions (fixed-point backends).

speps:
It's still branching inside a tight loop... :)

radarsat1:
I think the idea of optionally supporting clipping is not terrible, but it's true that it would add overhead.  I'm actually dealing with the old "branching in a tight loop" problem on another project.. apart from run-time code generation, which sounds difficult and messy, I can only think of a few possible solutions:

1. specialisation (checking booleans ahead of time and calling a customised loop
2. don't incorporate new features that require branching
3. deal with the overhead

The first solution can be somewhat alright, as long as the loop is expressed succinctly, but it can lead to bloated code very easily.

garyscavone:
I prefer not to add the extra overhead. Let the user add that if they want. I considered this a long time ago and decided against it.

PR #67, opened 2016-07-22
refactor: C++0x compatible exception semantics

antis81:
Add C++0x compatible exception semantics. This is probably not yet compatible accross compilers - ~~and not backwards compatibile as well~~ (**EDIT:** should be backwards compatible). Could someone review and check this, please? Would be nice to see this become cross-compatible with the "most common" compilers (vs/nmake, clang, gcc).

**CHECKS (Help wanted!)**
- [ ] Check with MSC (Microsoft) compiler
- [ ] Check with older (C++03 only) compiler (e.g. GCC 4.3)

---

This change is [Reviewable](https://reviewable.io/reviews/thestk/rtaudio/67)

### DIFF
--- a/RtAudio.cpp
+++ b/RtAudio.cpp
@@ -92,12 +92,12 @@ const unsigned int RtApi::SAMPLE_RATES[] = {
 //
 // *************************************************** //
 
-std::string RtAudio :: getVersion( void ) throw()
+std::string RtAudio :: getVersion( void )
 {
   return RTAUDIO_VERSION;
 }
 
-void RtAudio :: getCompiledApi( std::vector<RtAudio::Api> &apis ) throw()
+void RtAudio :: getCompiledApi( std::vector<RtAudio::Api> &apis )
 {
   apis.clear();
 
@@ -209,7 +209,7 @@ RtAudio :: RtAudio( RtAudio::Api api )
   throw( RtAudioError( errorText, RtAudioError::UNSPECIFIED ) );
 }
 
-RtAudio :: ~RtAudio() throw()
+RtAudio :: ~RtAudio()
 {
   if ( rtapi_ )
     delete rtapi_;
--- a/RtAudio.h
+++ b/RtAudio.h
@@ -214,22 +214,22 @@ class RtAudioError : public std::exception
   };
 
   //! The constructor.
-  RtAudioError( const std::string& message, Type type = RtAudioError::UNSPECIFIED ) throw() : message_(message), type_(type) {}
+  RtAudioError( const std::string& message, Type type = RtAudioError::UNSPECIFIED )  : message_(message), type_(type) {}
  
   //! The destructor.
-  virtual ~RtAudioError( void ) throw() {}
+  virtual ~RtAudioError( void ) {}
 
   //! Prints thrown error message to stderr.
-  virtual void printMessage( void ) const throw() { std::cerr << '\n' << message_ << "\n\n"; }
+  virtual void printMessage( void ) const { std::cerr << '\n' << message_ << "\n\n"; }
 
   //! Returns the thrown error message type.
-  virtual const Type& getType(void) const throw() { return type_; }
+  virtual const Type& getType(void) const { return type_; }
 
   //! Returns the thrown error message string.
-  virtual const std::string& getMessage(void) const throw() { return message_; }
+  virtual const std::string& getMessage(void) const { return message_; }
 
   //! Returns the thrown error message as a c-style string.
-  virtual const char* what( void ) const throw() { return message_.c_str(); }
+  virtual const char* what( void ) const { return message_.c_str(); }
 
  protected:
   std::string message_;
@@ -375,7 +375,7 @@ class RtAudio
   };
 
   //! A static function to determine the current RtAudio version.
-  static std::string getVersion( void ) throw();
+  static std::string getVersion( void );
 
   //! A static function to determine the available compiled audio APIs.
   /*!
@@ -383,7 +383,7 @@ class RtAudio
     the enumerated list values.  Note that there can be more than one
     API compiled for certain operating systems.
   */
-  static void getCompiledApi( std::vector<RtAudio::Api> &apis ) throw();
+  static void getCompiledApi( std::vector<RtAudio::Api> &apis );
 
   //! The class constructor.
   /*!
@@ -401,10 +401,10 @@ class RtAudio
     If a stream is running or open, it will be stopped and closed
     automatically.
   */
-  ~RtAudio() throw();
+  ~RtAudio();
 
   //! Returns the audio API specifier for the current instance of RtAudio.
-  RtAudio::Api getCurrentApi( void ) throw();
+  RtAudio::Api getCurrentApi( void );
 
   //! A public function that queries for the number of audio devices available.
   /*!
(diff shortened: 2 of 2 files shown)


doe300:
Compiles on Fedora 23 with both GCC 5.3.1 and Clang++ 3.7.0

antis81:
@doe300 Thanks!

Anyone having an older compiler available (e.g. gcc 4.3), with c++03 support only? Furthermore it's interesting to check with MSC (Microsoft) compiler.

xXSeraphimXx:
I could check it with MSC but I do not know how to enable the use of the older compiler (I'm using Visual Studio 2015)

antis81:
@xXSeraphimXx That's fine. Could you check, please?

xXSeraphimXx:
Anyway yes of course I'll do a general test. I'm very interested in RtAudio,  because all my further projects are build on this Api :D

antis81:
Same here. Just started to implement an API wrapper with it and stumbled upon the warnings. Mind to have a look over to [qTox RtAudio implementation](https://github.com/tux3/qTox/pull/3524)?

radarsat1:
Is there any really good reason to add noexcept here?  If throw() is deprecated, maybe we should just remove it and call it a day.  I don't think it can break anything to just remove throw(), even for old versions, can it?

radarsat1:
To be clear, the only reason I can see noexcept being useful is in templated code, but that's not the case here..

antis81:
>… I don't think it can break anything to just remove throw(), even for old versions, can it?

Yeah, should be safe to remove throw() - and it increases compatibility throughout compilers. :) It's been some time since I dealt with exceptions and I think error codes are the better option in general - especially for a cross-platform API. Don't know if and how that fits in RtAudio however without breaking backward compatibility.

radarsat1:
I also agree regarding exceptions, but I don't think they'll be removed from RtAudio without a major version bump, as you are right about it being a big change in terms of API.  In any case, they are a perfectly normal part of C++, even if not everyone likes to use them ;)

antis81:
@radarsat1 So do we agree, that removing `throw()` is fine?

radarsat1:
I am personally fine with it.  Will have to do some testing of course.

antis81:
@radarsat1 I replaced/removed the macro and removed any throw(). Could you check, please?

radarsat1:
With g++ 6.1.0, I get:

~~~
In file included from RtAudio.cpp:43:
./RtAudio.h:232:23: error: exception specification of overriding function is more lax than base version
  virtual const char* what( void ) const { return message_.c_str(); }
~~~

Also, your text editor seems to have added some weird line endings, e.g.:

~~~
-std::string RtAudio :: getVersion( void ) throw()
+std::string RtAudio :: getVersion( void )^M
~~~

radarsat1:
Reading up on `std::exception` a bit, it seems maybe RtAudioError should be derived from `std::runtime_error` instead, or something like that? not sure.  Probably quite some of its functionality could be replaced by a standard exception handler.

radarsat1:
Although it seems that simply changing the definition of `what()` to,

    virtual const char* what( void ) const  _GLIBCXX_NOEXCEPT { return message_.c_str(); }

and adding 

    #ifndef _GLIBCXX_NOEXCEPT
    #define _GLIBCXX_NOEXCEPT
    #endif

somewhere above the class definition seems to do the trick.

alexeysvrv:
@antis81 What do you think?

antis81:
It should be fine, even with MSC.

@radarsat1 The `^M` characters are carriage returns (CR => 13). This is standard on basically all Unix systems nowadays, including OSX, which has used LF (10) in the past. I'm not sure about Win10, but Windows used to end a line with CRLF (13, 10). Actually a lot of cross-platform projects seem to use the Unix format, as I haven't faced such things since years now… Unfortunately I cannot change that, because it would require to be set on a "per project" basis and I didn't setup a project for RtAudio. To help potential contributors with this particular topic, I'd suggest to convert files to consequently use CR line endings and - if possible - UTF-8 encoding. Even VS should be able to set this encoding today.

DouglasHeriot:
@antis81 you are incorrect about CR vs LF.
LF is the standard on Unix (including OS X), and Windows uses CRLF. Old Mac OS (≤9) used CR.
The ^M showing up means you’re using a Unix platform expecting only LF, but windows has put in both CR and LF.
rtaudio already uses LF everywhere, as is the standard for unix projects. You can configure Windows git clients to convert LF to CRLF on checkout, and back to LF on commit. Or just make Visual Studio use LF only.

antis81:
@DouglasHeriot Erm… I'm on Linux… AFAIK it uses LF, but I'll re-check that… (Confirmed!) :)

>rtaudio already uses LF everywhere…

Whilst you are right about the LF on Unix systems (sorry, I mixed that up), the `^M` definitely marks a CR character. If looking more closely, you'll find the character gets "deleted" in the Git diff.

Now I searched a bit for ways to check that, and here we go:

```bash
$ file RtAudio.cpp 
RtAudio.cpp: C source, ASCII text, with CRLF line terminators
```

Obviously that file (on current master branch) is CRLF encoded. To solve that, it can be converted to LF via `dos2unix`:

```bash
$ dos2unix RtAudio.cpp
$ file RtAudio.cpp
RtAudio.cpp: C source, ASCII text
```

If that isn't enough confusion, the header file already is in the right format. :)

antis81:
@DouglasHeriot Please have a look in recent PR #73, thanks!

radarsat1:
Yeah, I see now that RtAudio.cpp uses CRLF and RtAudio.h uses LF line endings, weird.  Well, let's continue that discussion in #73 and concentrate on the exceptions thing here.  What do you think about adding `_GLIBCXX_NOEXCEPT`?

garyscavone:
As far as I see, all this PR does is remove the "throw()" statements from the function declarations in which no exception can be thrown. What is the goal here?

radarsat1:
Short version: It's because throw() is deprecated in newer versions of C++.

Long version: https://akrzemi1.wordpress.com/2011/06/10/using-noexcept/

garyscavone:
OK, seems like throw() is deprecated in newer versions. You can go ahead and commit this.

radarsat1:
Whoops.. something broke hold on

radarsat1:
Fixed I hope.

garyscavone:
I just tried it on OS-X 10.12 and I get the following errors:

In file included from RtAudio.cpp:43:
./RtAudio.h:224:11: error: exception specification of overriding function is more lax than
      base version
  virtual ~RtAudioError( void ) _GLIBCXX_USE_NOEXCEPT {}
          ^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/exception:95:13: note: 
      overridden virtual function is here
    virtual ~exception() _NOEXCEPT;
            ^
In file included from RtAudio.cpp:43:
./RtAudio.h:236:23: error: exception specification of overriding function is more lax than
      base version
  virtual const char* what( void ) const _GLIBCXX_USE_NOEXCEPT { return message_.c_str(); }
                      ^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/exception:96:25: note: 
      overridden virtual function is here
    virtual const char* what() const _NOEXCEPT;
                        ^
2 errors generated.

—gary

radarsat1:
Shoot. Added a new attempt.

radarsat1:
Tomorrow I might check whether this kludgy noexcept stuff can be removed by simply not overriding what() etc.. since I think std::runtime_error actually already does most of what RtMidiError is doing.

garyscavone:
That fixed the compile problems!

PR #68, opened 2016-07-26
a JACK-specific flag to disable automatic connection of ports

jpcima:
I propose a patch to implement what the title says.

RtAudio will connect the Jack ports as the program initializes, which is not always the desired behavior, and it lacks a programmatic way to disable this.

On the typical Linux softsynth, it is expected to begin unconnected and let the user manage connections externally, either manually or using the patch bay.

The patch does not modify existing behavior, instead it introduces a stream flag to realise my desired behavior. I have tried the functionality on a softsynth I am currently developing.

### DIFF
--- a/RtAudio.cpp
+++ b/RtAudio.cpp
@@ -1943,7 +1943,7 @@ struct JackHandle {
 static void jackSilentError( const char * ) {};
 
 RtApiJack :: RtApiJack()
-{
+    :shouldAutoconnect_(true) {
   // Nothing to do here.
 #if !defined(__RTAUDIO_DEBUG__)
   // Turn off Jack's internal error reporting.
@@ -2354,6 +2354,8 @@ bool RtApiJack :: probeDeviceOpen( unsigned int device, StreamMode mode, unsigne
   // here.
   if ( stream_.doConvertBuffer[mode] ) setConvertInfo( mode, 0 );
 
+  if ( options && options->flags & RTAUDIO_JACK_DONT_CONNECT ) shouldAutoconnect_ = false;
+
   return SUCCESS;
 
  error:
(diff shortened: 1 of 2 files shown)


garyscavone:
Let me make sure I understand the desired behaviour here. If you start a stream running without a connection being made, does the callback handler still get called? And if so, I guess it simply returns unless there is valid data to process? I assume if an external program establishes a connection, then valid data will be present?

radarsat1:
For some context, afaik the "ports should automatically connect or not" issue is a long-running disagreement among Jack users ;)  The kind of thing that should be an option, so probably this contribution would be useful for a certain subset of Jack users.  Possibly in a future version we could come up with a more general "backend-specific options" interface, I dunno, but adding a new flag like this seems like a reasonable approach for now.  I would generalize it, taking out "Jack" from the name, and just call it `RTAUDIO_NO_AUTOCONNECT` or something like that -- maybe it could be useful for other backends.

garyscavone:
I'm not sure how generalizable this would be. For most APIs, the expectations is that you have a connection between two ports _before_ starting a stream running. I'd have to look into the details of how Jack is working, but it seems like what we initially do in the "startStream()" function (jack_activate) does not start the flow of audio samples but rather that happens after "jack_connect" is invoked.

radarsat1:
Ah that's true, regarding "startStream".. in general for jack the "stream running or not" state is independent from whether it is actually connected to an output.  Some users prefer to use an external tool to maintain their Jack connections as a "session" instead of letting the jack client program control it.

garyscavone:
I'm going to go ahead with this. Perhaps in the future, there may be other cases to which it can be generalized.

jpcima:
Being a bit late to the topic I don't have much to add which has not already been said.
The input ports of Jack receive the sum of all ports connected to them, so naturally it gets zero in case of no connection, and the audio callback always runs at the normal rate.

I think that having a choice of whether to connect is important for practical musical programs, because otherwise it nullifies the usefulness of the connection manager. The discussion about jack is probably not really about whether the client should connect or not, but which of the behaviors should be default.

For information I observed the puredata project has taken notice of the same issue in a relatively recent git version and has added a flag to provide the relevant choice.

    -jackconnect     -- automatically connect pd to the JACK graph [default]

PR #73, opened 2016-08-15
refactor: convert line endings to unix format (LF)

antis81:
Converts line endings in RtAudio.cpp to cross-platform (Unix => LF) format.

**Important note:**
After merge/rebase onto master, probably every open PR will be in conflict state and has to be rebased onto master. Can we send a notification mail to all developers?


speps:
I don't see the point, that should be done with Git options really.

radarsat1:
I think this should be fixed, but I don't think this patch should be applied until immediately before a release.

garyscavone:
I don't think RtAudio.cpp used to have CR line endings. Perhaps somewhere along the line, we made a commit that added them?

radarsat1:
Regardless, this sorely needs to be done.. I think my last commit added one LF and now emacs loads the whole file with LF and the CRs are displayed.  Very weird.  (And sorry about that, I'll fix it.)

Instead of merging this however, I suggest simply running the whole source through dos2unix immediately before the next release.  There are relatively few pull requests right now so it's a good time.

garyscavone:
Fixed in new release.

PR #98, opened 2017-09-30
add rtaudio c api wrappers

zserge:


radarsat1:
Hi, the code looks good, just a couple of problems with the configuraiton.

1. Remove the change to AC_CONFIG_SRCDIR, it only takes a single file in the srcdir as its argument.
2. Add rtaudio_c.cpp to the RtAudio.cpp line in Make.am instead
3. with -Wall, compiler complains about unused parameter "errcb" in rtaudio_open_stream.

Also, nit-pick, but I'd prefer if you use spaces instead of tabs.  (To avoid any large patches later down the line like the recent dos2unix conversion..)

Thanks!

zserge:
@radarsat1 Thanks for reviewing, I'll update it now. Out of curiosity, do you have any coding style so I could set up my `cmake-format`? Cause now I used its default formatting settings.

radarsat1:
Not really, but most of the code is spaces.  I don't know cmake-format, although I have used other tools in the past like "indent".  In any case there is no enforced coding style other than what you can infer, I just thought it might be nice to avoid the tab-space debate and at least specify that much ;)

zserge:
I've updated the pull request. Also, looks like spaces have been used previously, with a few exceptions of deep indentation where a mixture of both was used. I've updated my `.clang-format` config for this project to never use tabs, hopefully it helps.

rtaudio_c.cpp
```diff
@@ -126,6 +126,7 @@ int rtaudio_open_stream(rtaudio_t audio,
 			unsigned int *buffer_frames, rtaudio_cb_t cb,
 			void *userdata, rtaudio_stream_options_t *options,
 			rtaudio_error_cb_t errcb) {
+  (void)errcb;
```
radarsat1: Better to just put `rtaudio_error_cb_t /*errcb*/` in the parameter list.  If `errcb` is unnamed then gcc doesn't complain.

zserge:
As for the `errcb` - I can't really use it because in C++ API there is no userdata or any other `void*`-like context passed into the error callback, which makes it impossible for me to wrap into C API. If you ever decide to change the C++ API - I'd be happy to update the C bindings to use `errcb` as expected.

radarsat1:
Understood regarding the error callback thanks for pointing it out.

If you just fix up the fix-up (comment-out cberr), and squash it down to one commit, I will push it.

rtaudio_c.cpp
zserge: Sorry, I have to argue with this. I don't think it's valid C99 (it does work in C++ only), while casting unused parameters to void is a portable and a rather common solution.

```
$ gcc -std=c99 -Wall -Wextra -pedantic -c x.c
x.c: In function ‘foo’:
x.c:1:1: error: parameter name omitted
 int foo(int) { return 0; }
```

zserge:
@radarsat1 Thanks for the feedback. I've commented on your review note about `errcb`. Also I think Github allows doing squash merges without making me do squash ("Confirm squash and merge" form the green merge button dropdown list).

zserge:
@radarsat1 Nevermind. I somehow forgot that my C API bindings are actually C++, not C. I've updated the pull request.

rtaudio_c.cpp
radarsat1: Huh. TIL.

radarsat1:
Ah, fair enough I usually merge from the command line so I didn't know about the web interface for squashing ;)

zserge:
@radarsat1 Would you be interested in Go bindings by any chance? I was hoping to put them into official repo to let people use RtAudio in Go (now everyone uses portaudio, which is really a pain IMHO).
Go bindings would add one file (`rtaudio.go`) into your root directory (because Go source should ideally be next to the C/C++ files it uses). I'd probably add another file (`rtaudio_test.go`) for tests/examples.
This should allow people to use RtAudio by just importing `import "github.com/thestk/rtaudio"` line in their Go code.

So if you find adding Go bindings benefitial - please let me know, I will polish my code and add documentation/tests and make a pull request then. Currently it looks like this: https://github.com/naivesound/audio/blob/master/rtaudio/rtaudio.go

The same question also applies to RtMidi.

Thanks for your collaboration!

radarsat1:
I think it would be cool personally.  Maybe @garyscavone has an opinion.  However, I would put any bindings in the "contrib" directory so that things remain clear and organized.

zserge:
I totally understand the preference to separate bindings from the core library code, and I wish there was a way to do so!
But Go is a very opinionated language, and their C bindings system (`cgo`) works in such a way that it automatically compiles all C/C++ code _inside_ the directory with the Go sources and not anywhere else. Of course, it might be possible to build RtAudio as a static library before building Go code, provide path to that library to `cgo` etc, but that would force every library user to do this steps during the build, which is a lot of inconvenience. So to keep the bindings developer-friendly I'm afraid the Go sources should be stored next to the C/C++ sources.

radarsat1:
That is unfortunate and weird.  I don't know Go, but with Python the bindings are just a separate library that you compile and link against the .so.  Another option would be to investigate SWIG and hit more than one target at a time.  (I know, not everyone's favourite solution, neither is mine, but it is as powerful as it is annoying to use..)

Anyways, if you've got them ready, maybe make an issue or PR to discuss this. I'm curious how you would go about writing the Go bindings if it was a separate repository.  If it's really *that* inconvenient, it sounds like a Go problem rather than an RtMidi problem.  My inclination is to insist on making it a separate project in that case, but I would see what @garyscavone thinks.  Mainly I am afraid of having something unfamiliar to test/maintain.

garyscavone:
I also have no experience with Go and I’m not in a good position to start learning about it for a while. We have the same support issue with things like cmake. At this point, I will defer to @radarsat1.

PR #99, opened 2017-10-03
Add go bindings

zserge:
Go bindings are two files, one is actual implementation using cgo, another is a test file with examples. Examples are rather simple (list devices and play a sine wave), but since they are built automatically during the `go test` process, they will help to detect problems like broken API early.
I've also checked my bindings with lint/vet and goreportcard seems to be happy with the code, too: https://goreportcard.com/report/github.com/zserge/rtaudio


radarsat1:
I'm sorry to insist but I really think it would be nice to put this in contrib instead of the main directory.  As such I experimented with a few ways and finally found one: creating stub files.  Please check my repository https://github.com/radarsat1/rtaudio/commit/30ac05af567fd78070a2ebe6df5c4c46a8720dda where I put it on master (since go doesn't know how to use branches)

With this file `testrtaudio.go`,

```go
package main

import (
	"log"
    "github.com/radarsat1/rtaudio/contrib/go/rtaudio"
)

func DoExampleCompiledAPI() {
	log.Println("RtAudio version: ", rtaudio.Version())
	for _, api := range rtaudio.CompiledAPI() {
		log.Println("Compiled API: ", api)
	}
}

func main() {
    DoExampleCompiledAPI()
}
```

I was able to run the following commands with an empty GOPATH directory, and they succeed without error:

```
~/projects/tmp
$ go get github.com/radarsat1/rtaudio/contrib/go/rtaudio

~/projects/tmp
$ go build testrtaudio.go 

~/projects/tmp
$ ./testrtaudio 
2017/10/03 12:15:10 RtAudio version:  5.0.0
2017/10/03 12:15:10 Compiled API:  alsa
```

zserge:
@radarsat1 Hm.. That's a great idea, haven't thought of it. Although I would separate `*.cpp` from `*.h` files to avoid including C++ code into `cgo` generate files. They might work, but some compilers are likely to be unhappy, because only C code is officially supported inside `cgo`.

Would you like me to update the PR?

radarsat1:
I'm not sure I understand. I thought the point here was to include the RtAudio.cpp code inside the bindings so that everything can be installed in one "go" command.  It seems there are two possibilities,

1. go/cgo compiles the cpp code, and it is included directly in the bindings, in which case the stub files work (empirically this "works for me")
2. C++ code cannot be directly included in the bindings, so RtAudio.cpp must be compiled separately.

In case (1) this solution seems to work.  In case (2), RtAudio would be considered an "already installed library" which you are compiling bindings against.  You can put any binding code in the contrib directory, and just need to `#include rtaudio_c.h`, leaving paths up to the system.  (CGO_CFLAGS etc.)  So (2) seems to conflict with your initial assertion that the binding code must be alongside RtAudio.cpp and rtaudio_c.cpp.

What am I missing?

zserge:
I'm now trying the same approach with RtMidi and it seems to work.

I have one `rtmidi_stub.h` that gets included from the `cgo` block inside the `rtmidi.go`:

```c
#include "../../../rtmidi_c.h"
```
It is a valid C99 code and it's safe to include into all C files generated by `cgo` during the build.

I have another file, `rtmidi_stub.cpp` that simply is located in the same `contrib/go/rtmidi` directory, but is not referenced from the Go source:

```c
#include "../../../RtMidi.h"

#include "../../../RtMidi.cpp"
#include "../../../rtmidi_c.cpp"
```

`rtmidi_stub.cpp` is compiled during the `go build`, because it's located in the same package directory. But it's not included in `cgo` block, so it's not copied into the genrated `*.c` sources.

radarsat1:
Oh I see, but that's what I did also, right?  My `rtaudio_stub.h` only includes `"../../../rtaudio_c.h"` as well.  Maybe I misunderstood your previous comment.

zserge:
@radarsat1 Oh, my bad, sorry. I have seen all four files included in `rtaudio_stub.cpp` and mislooked the header file next to it, I though you managed to put it all in one stub file only. So I have ended up doing the same in my repo as well. Feel free to merge either of them, then.

And, if this approach looks good to you - I'd like to do a pull request to RtMidi now, too.

radarsat1:
Very cool, glad we found a compromise ;)

PR #101, opened 2017-10-06
Use "using namespace std" for roundf instead of std::roundf.

radarsat1:
@zserge can you check if this helps with your problem?


zserge:
@radarsat1 Yes, it does the trick!

zserge:
@radarsat1 Could you please merge it, so I could make another pull request with Go-related changes on top of this one?

PR #102, opened 2017-10-06
go binding fixes

zserge:
- Use static build flag for windows to work around mingw issues (MinGW on windows otherwise has duplicate symbols because it tries both static and dynamic libraries at the same time)
- Use array sizes adjusted for 32-bit systems (otherwise on 32-bit systems there would be build errors because array size can theoretically exceed 32-bit address space).


zserge:
This together with #101 should fix Go builds on both 64 and 32-bit Windows.

radarsat1:
Merged, but now I look at the history and realized I should have asked you to rebase it on #101. Oh well.

radarsat1:
New commit is 2e5704f2a69a3195fa06934dcf33dabba9777ecd

PR #104, opened 2017-10-09
Debian fixes

umlaeute:
a PR with minor issues we found during preparing the Debian package


umlaeute:
after reading https://github.com/thestk/rtmidi/issues/131, i thin kthat the second patch 
a8fcb15 is not quite correct. it should include something along the lines of `@api@`

PR #106, opened 2017-10-16
fixed cmake install error : rtaudio.pc error

bejuryu:
I fixed the rtaudio.pc file error when I build and install using cmake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -125,6 +125,8 @@ if (BUILD_TESTING)
     add_subdirectory(tests)
 endif (BUILD_TESTING)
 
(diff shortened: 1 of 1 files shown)


radarsat1:
Thanks!  It seems though that this change doesn't properly replace the variables in the file `rtaudio.pc.in`.  The resulting `rtaudio.pc` has empty Version and Requires fields and is missing the API define in Cflags.

radarsat1:
I merged this for now but it would be nice to address the further problems with missing variables if possible.

radarsat1:
This was fixed in adefaebdbcc1c3137b1c9ce8d995ef25564c7fc8

PR #107, opened 2017-10-16
Support Debian/kFreeBSD and Debian/Hurd

umlaeute:
i finally got around making the build-system work with Debian/kFreeBSD and Debian/Hurd.
(i'm aware that these are probably not the most popular systems to do any serious realtime and/or audio work; but you never know...)

note that the patch is not very thoroughly tested.
It's simple, so i don't think it will break anything; but the actual builds on these platforms have only been scheduled yet, and not yet executed (well `kfreebsd-i386` has already succeeded); see [Debian's buildd](https://buildd.debian.org/status/package.php?p=rtaudio&suite=experimental) for the current status.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -246,6 +246,11 @@ AS_CASE([$host],
     LIBS="-ldsound -lwinmm $LIBS"
   ])
   LIBS="-lole32 $LIBS"
+  ],
+  [*-*-kfreebsd*-gnu | *-*-gnu*],[
+  AS_IF([test "x$api" = "x" ], [
+    AC_MSG_ERROR(["$host" system has no default backend])
+  ])
   ],[
   AC_MSG_RESULT([none])
   # Default case for unknown realtime systems.


radarsat1:
> "$host" system has no default backend

but is there any point specifically matching on unsupported systems?  why not just make this an AC_MSG_WARN instead and make it the exception if no known platforms are matched?

umlaeute:
I have no idea why the RtAudio developers insisted on `configure` to fail if `host` returns a value they haven't though of.

I'm not sure what you mean with "make it the exception if no known platforms are matched", but I'm pretty sure that making assumptions about "known platforms" defeats the idea of autoconf. Esp, `configure` shouldn't fail if it encounters an unknown platform.

most likely, the fix can be as simple as:

~~~diff
-AC_MSG_ERROR([Unknown system type for realtime support!])
+AS_IF([test "x$api" = "x" ], [
+   AC_MSG_ERROR(["$host" system has no default backend])
+])
~~~

(and drop the special handling of kFreeBSD and GNU/hurd)

radarsat1:
I agree, the CASE should be only for selecting the default API is none is suggested by `$host`, then it should only fail if no API is selected.  Otherwise it should allow users to specify whatever API they prefer without checking the platform.

radarsat1:
I believe this should be fixed in adefaebdbcc1c3137b1c9ce8d995ef25564c7fc8, please re-open if necessary.

Issue #69, opened 2016-07-29
WASAPI: when using non-default sample rates, there are audio artifacts


paulmasri:
_Affects RtAudio 4.1.2_

There appears to be very crude resampling of waveforms when the selected sample rate for the output stream doesn't match the default sample rate of the device. I have tested this on Windows 7, 8 & 10.

Here is an example waveform, captured via loopback, where the selected sample rate is 44100Hz and the default sample rate is 48000Hz. The red arrows indicate the "extra" samples that appear to have been added to achieve resampling and which result in significant audio artifacts.

![screen shot 2016-07-29 at 15 20 34](https://cloud.githubusercontent.com/assets/496858/17251305/0effed6c-55a0-11e6-8c02-f387e51e13ec.png)

Possibly related to #22 ?

garyscavone:
Yes, this is a significant problem with the WASAPI implementation.  If I remember correctly, it resamples to nearest neighbour with truncation and at least could be improved with some linear interpolation. It would probably be better if the WASAPI implementation was in a development branch, not in the release branch, until this problem is fixed.

paulmasri:
Thanks for all the awesome work you've done over the years @garyscavone. Please don't pull WASAPI... I'm relying on it for something that's going to production! :-o My workaround is to find out the default device's preferred sample rate and to use that (and do linear interpolation resampling within my app).

MarcusTomlinson:
Damn, turns out I haven't properly fixed this issue yet. Turns out my interpolation logic was overrunning its buffers. Sorry about that!

radarsat1:
Is it not possible to fix the buffer overrun without reverting the whole thing?  I'll merge your revert if necessary but are you sure you were completely on the wrong track?

radarsat1:
Hm now that I'm looking at the code it seems to me that `WasapiBuffer` and `convertBufferWasapi` could be tested on their own without the need for WASAPI itself.  Can you suggest a way to test / detect the issue you are reporting?

MarcusTomlinson:
It's not that I was on the wrong track really, I just felt that in the interim it'd be way better to leave users with a WASAPI API that actually worked while I investigate a fix. (The artefacts my "fix" introduced are far worse than the artefacts we had)

If you're interested in looking into it, the problem is with my retrieval of the **toSample**:
I'm indexing **toSample** as `[ frameInSample + channelCount + channel ]`
(Where `+ channelCount` is a frame jump to the sample following **frameInSample**).
[...]

PR #119, opened 2017-12-23
Fix Issue #69 - Allow non-default sample rates in WASAPI

MarcusTomlinson:
As WASAPI itself does not provide support for sample rates other than the current system default, I've introduced the Microsoft MFT (Media Foundation Transform) library into RtApiWasapi in order to allow users to specify the sample rate they wish to use.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -178,7 +178,7 @@ endif()
 # WASAPI
 if (RTAUDIO_API_WASAPI)
   set(NEED_WIN32LIBS ON)
-  list(APPEND LINKLIBS uuid ksuser)
+  list(APPEND LINKLIBS ksuser mfplat mfuuid wmcodecdspuuid)
   list(APPEND API_DEFS "-D__WINDOWS_WASAPI__")
   list(APPEND API_LIST "wasapi")
 endif()
--- a/RtAudio.cpp
+++ b/RtAudio.cpp
@@ -3700,6 +3700,7 @@ static const char* getAsioErrorString( ASIOError result )
 #include <Wmcodecdsp.h>
 
 #pragma comment( lib, "mfplat.lib" )
+#pragma comment( lib, "mfuuid.lib" )
 #pragma comment( lib, "wmcodecdspuuid" )
 
 //=============================================================================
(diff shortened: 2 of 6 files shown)


paulmasri:
@MarcusTomlinson FWIW Travis is complaining about 94dbbb2 (in case you didn't see it)

MarcusTomlinson:
Hmm, I've only ever built RtAudio with VC++ (so all good there of course). Looks like I'll need to install MinGW and try to reproduce the build failure locally.

KnappSas:
Hey @MarcusTomlinson .. Will there be some progress on this issue?

MarcusTomlinson:
@KnappSas, unfortunately I've been swamped these last 2 months. I was actually thinking of asking if anyone was able and willing to assist me on this one. All builds fine on VS, I just haven't worked with MinGW much, so I'd need a good couple days to get set up and fix it myself.

radarsat1:
It appears that the WASAPI headers in MingW define the same structs. (ksmedia.h and stfmif.h both define `struct tagTIMECODE_SAMPLE`

radarsat1:
It could be a bug in MingW.  I haven't found anything in a search on the topic.
Anyways, if you want to reproduce I suggest using a Linux system or Docker instead of installing MingW on Windows, since that is the environment used by Travis.  It is not really necessary to support WASAPI on MingW/Linux, it is only used for testing on Travis, however it would be a shame to have to disable this configuration as it helps for testing that codepath.

KnappSas:
@MarcusTomlinson On my local Windows 8.1 I wasn't able to build your stuff I'm afraid. 

> 1>c:\program files (x86)\windows kits\8.1\include\um\strmif.h(13490): error C2011: 'tagTIMECODE_SAMPLE': 'struct' type redefinition
1>c:\program files (x86)\windows kits\8.1\include\shared\ksmedia.h(5880): note: see declaration of 'tagTIMECODE_SAMPLE'
1>c:\program files (x86)\windows kits\8.1\include\um\ddraw.h(713): error C2011: '_DDPIXELFORMAT': 'struct' type redefinition
1>c:\program files (x86)\windows kits\8.1\include\shared\ksmedia.h(6361): note: see declaration of '_DDPIXELFORMAT'
1>c:\program files (x86)\windows kits\8.1\include\um\ddraw.h(2259): error C2079: '_DDSURFACEDESC::ddpfPixelFormat' uses undefined struct '_DDPIXELFORMAT'
1>c:\program files (x86)\windows kits\8.1\include\um\ddraw.h(2302): error C2079: '_DDSURFACEDESC2::ddpfPixelFormat' uses undefined struct '_DDPIXELFORMAT'

Any hint?

radarsat1:
Oh interesting, that's very similar to the error from MingW, so maybe there are some headers that cannot go together here?  Or is there a global header that is supposed to be used instead?

KnappSas:
What about this https://msdn.microsoft.com/en-us/library/windows/desktop/dd389294(v=vs.85).aspx ? 

>  [MFPlay is available for use in the operating systems specified in the Requirements section. It may be altered or unavailable in subsequent versions. ]

..and requirements say Windows 7.

MarcusTomlinson:
This branch was building and running just fine in Visual Studio 2018 on my Windows 10 machine.

radarsat1:
I believe you, but "works for me" is not really good enough, especially if it breaks things for Windows <10.  We need to figure out how to correctly detect if this feature is available and only turn it on under those circumstances.

MarcusTomlinson:
Haha

KnappSas:
I actually just build it via trial & error. I moved all Media Foundation includes up above audioclient.h. Unfortunately I now don't get any sound in the app where I'm using RtAudio :(

radarsat1:
I just tried what you describe on MingW and it seems to help but I get,

```
RtAudio.cpp:4051:3: error: ‘IMFTransform’ does not name a type
```

so I added `#include <mftransform.h>` and then I get,

```
RtAudio.cpp:4054:3: error: ‘IWMResamplerProps’ does not name a type
```

but as far as I can tell [from here](https://msdn.microsoft.com/en-us/library/windows/desktop/ff819250(v=vs.85).aspx) this should be defined in `wmcodecdsp.h` which is definitely included.

So it seems maybe a way to detect if the necessary feature is available is to check defines in that include.
I found a Windows 7 copy of that file [here](https://searchcode.com/codesearch/view/66548741/)

and it seems we can check `#ifdef __IWMResamplerProps_FWD_DEFINED__`

radarsat1:
Can we gate all the using WasapiResampler with,

```
#ifdef __WASAPI_ALLOW_NON_DEFAULT_SAMPLE_RATES__                                        
#endif
```

and define immediately after the includes,
```
#ifdef __IWMResamplerProps_FWD_DEFINED__                                                 
#define __WASAPI_ALLOW_NON_DEFAULT_SAMPLE_RATES__                                        
#endif 
```

radarsat1:
Admittedly it's weird that IWMResamplerProps is not available in that header, since the doc page clearly says it's available since Windows XP, but I guess the Windows Media Foundation headers in MingW are not quite up to scratch, which is not a great reason to not include this patch.  I would just like to gate it with that preprocessor define so that the rest of the WASAPI support can be tested.

paulmasri:
Just in case it's relevant, which versions of MinGW are each of you using? I'm curious why it compiles on one Win10 system and not another. I have definitely experienced issues with some versions of MinGW (when compiling my Qt app). Maybe this might help work out whether it's a MinGW issue or a repo issue.

radarsat1:
I'm testing with mingw from Ubuntu 17.04:

```
$ x86_64-w64-mingw32-gcc --version
x86_64-w64-mingw32-gcc (GCC) 6.2.1 20161119
```

Although I think Travis uses an older version of Ubuntu, 14.04

KnappSas:
So the audio issue for me was fixed. I opened a pull request for this in Marcus repository.

KnappSas:
Hello,
I also tested it on Ubuntu 16.04 with the same result:
`x86_64-w64-mingw32-gcc (GCC) 5.3.1 20160211`

So would we consider the proposal from @radarsat1 using the preprocessor switches?

MarcusTomlinson:
Thanks for the pointers guys. And sorry it took me so long to get back around to this!

I've managed to fix the build now. It was a combination of reordering includes, adding missing linker flags, and dodging some missing definitions in the mingw headers.

Looks all good to me now.

MarcusTomlinson:
@radarsat1 you happy to merge this now?

radarsat1:
Seems checks are passing so probably yes.  I haven't had time to review rtaudio/rtmidi stuff lately, so I'll have to get to it when I can.  Thanks for the contribution!

PR #122, opened 2018-01-02
Support for Genode Audio_out and Audio_in

ehmry:
I patched in Genode support over a year ago for a ChucK port. The patch still applies cleanly so I thought I'd make a PR.

http://genode.org/


radarsat1:
I don't think there's too much interest after this has been here for the whole year.  Would you like to maintain it as a fork perhaps?

ehmry:
Yes, that is fine. The API is stable so its not a hassle.

Issue #116, opened 2017-11-25
PulseAudio is available for non-linux systems


falkTX:
The current code makes it sound like pulseaudio is a linux specific API.
But I was able to build rtaudio with pulseaudio enabled for freebsd and gnu-hurd.

Would be nice to rename LINUX_PULSE stuff to UNIX_PULSE.

radarsat1:
While I don't disagree, I'm not sure I see the advantage of possibly breaking configuration / build system code for other people's projects for such cosmetic reasons.  Let's put this on the pile along with C++11 of "things we'll break when we decide to break everything"

falkTX:
Makes sense..
I personally renamed the macro to make it less confusing, but it will be a breaking change for sure.

tay10r:
If we're talking about renaming the macros, I would go so far as to say drop the operating system names entirely. I would also like to suggest removing the underscores in the beginning of the identifiers and making the macros more descriptive. Here's what I would suggest for the header and build systems.

```c++
#define RTAUDIO_WITH_JACK
#define RTAUDIO_WITH_ALSA
#define RTAUDIO_WITH_CORE
#define RTAUDIO_WITH_PULSE
#define RTAUDIO_WITH_WASAPI
#define RTAUDIO_WITH_ASIO
#define RTAUDIO_WITH_DS // Or RTAUDIO_WITH_DIRECTSOUND
```

Or something close to that.

Parakleta:
FreeBSD needs to use /usr/local/lib or /usr/local/include to access the pulseaudio libraries.

ntonnaett:
With #313 API defines shouldn't be necessary anymore. So we would only have to change RtAudio's build systems.

falkTX:
I saw that PR changes just now, doesn't seem complete.

ntonnaett:
What is missing?

falkTX:
I left a few comments in the code, hopefully github is able to show them this time...

ntonnaett:
I've actually meant #308 but your point there is true for both.

radarsat1:
Closing as solved with #308 and #313.

Issue #110, opened 2017-10-28
Allow build on non-linux architectures


miramikes:
Hi,

following patch allow rtaudio to be build on non-linux archs as hurd-i386 kfreebsd-amd64 kfreebsd-i386.
You might be interested to apply it upstream.

mira
[0001-kfreebsd-hurd.txt](https://github.com/thestk/rtaudio/files/1424301/0001-kfreebsd-hurd.txt)

radarsat1:
This should now be resolved as of adefaebdbcc1c3137b1c9ce8d995ef25564c7fc8

PR #135, opened 2018-05-01
autoconf: separate OS defaults and audio API testing

radarsat1:
This is an idea for reorganizing the way APIs are detected at configure time.

Instead of wrapping up OS detection with expected APIs, and giving errors for unknown OS, first the `with` options are considered, then *only if there are none specified*, the OS is detected and some defaults are suggested.  Then, any APIs on that list are tested regarding found libraries.  APIs specified by `with` options are considered 'required' (error generated if not found), otherwise they are optional (not-found APIs are ignored).

This allows unknown OSs to still specify an API using `with` options, and makes it easier to manage which OSs are associated with which APIs, and also avoids repeating detection code for multiple OSs that support a single API.

This addresses #107, #110, #116.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,12 +27,6 @@ matrix:
     env: HOST="" API="oss"
     compiler: clang
   - os: linux
-    env: HOST="--host=i686-w64-mingw32" API="winmm" CPPFLAGS="-Wno-unused-function"
-    compiler: gcc
-  - os: linux
-    env: HOST="--host=x86_64-w64-mingw32" API="winmm" CPPFLAGS="-Wno-unused-function"
-    compiler: gcc
-  - os: linux
     env: HOST="--host=i686-w64-mingw32" API="dsound" CPPFLAGS="-Wno-unused-function"
     compiler: gcc
   - os: linux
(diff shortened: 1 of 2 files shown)


radarsat1:
Sorry, not really #116.

garyscavone:
I see mention in github that a few of the Travis checks failed (oss, winmm, dsound … perhaps problems with the Travis script or the new changes?). Also, I tried on an OS-X machine with the “—with-ds” flag and no errors were reported (it chose CoreAudio). Perhaps that is the way it was intended to work?

radarsat1:
Yup, seems like some legitimate failures esp regarding OSS on linux, I'll check it out

radarsat1:
ps., regarding your OS X test, no, normally it would be better give an error in that case, except the Windows APIs don't seem to include actual tests (that is, don't on master, either).  We could add AC_CHECK_LIB for DS and WASAPI perhaps.

radarsat1:
Looking better now.  I didn't realize libossaudio is not needed on Linux, but is needed on the BSDs.  Also apparently I have been for quite a while testing a non-existent Windows configuration, oops.  (It was using the default DirectSound before, but now the default is to enable all known systems for a given OS, which raised an error due to `-Wall` on the ASIO code.)

garyscavone:
Yes, good by me! Thanks!

radarsat1:
Excellent ;)

PR #136, opened 2018-05-07
API names, and conversion from/to API identifier

jpcima:
Proposing an API to be able to get API names, and API identifiers according to names.

The purpose is to help developing programs which would present an enumeration of APIs as settings for users, or to allow designating an API by passing it as string on the command line.
If this suggestion is accepted, I wish to implement an identical proposal in RtMidi as well.

I add a unit test. This one would help to ensure nothing is forgotten if a new API were to be added at a future time.


trebmuh:
Sounds great!

radarsat1:
Personally I think it's useful although I would make the implementation a bit simpler using an std;:map.  I'm not sure it's necessary to make it case insensitive.

I think the downside to doing this is that now *we* choose what strings people use to refer to the APIs, so if we call it "core" or "CoreAudio" or "CoreAudio Mac OS X", now user programs might be tempted to hard-code these values and cause problems if they change in the future instead of using the enum identifiers which will cause compilation errors.  In that sense it is "safer" to force user programs to decide which strings they will use themselves.  But, I agree that it makes user programs less flexible as now they have to provide strings for each identifier and new APIs won't be automatically supported.  So there are pros and cons imho.

jpcima:
I think you raise some good points, and I can imagine some solutions to manage these inconvienents.

> Personally I think it's useful although I would make the implementation a bit simpler using an std;:map. I'm not sure it's necessary to make it case insensitive.

I considered case-insentivity for such reasons: it doesn't create ambiguities; and secondly it's more familiar in the CLI world, where the name comes from user's inputs, and for instance JACK is usually referred as lower-case in this setting everywhere as opposed to the official naming.

Also, there are many software (libao, mpv, etc) which favorize a lower-case naming.

> I think the downside to doing this is that now we choose what strings people use to refer to the APIs, so if we call it "core" or "CoreAudio" or "CoreAudio Mac OS X", now user programs might be tempted to hard-code these values and cause problems

I think of a solution like Common Lisps package naming.

What if an API can have
- list of nicknames : ie. "PulseAudio", "Pulse"
- short name, aka the first nickname
- a display name

In such a way, a lookup which has worked at one point can be allowed to work forever.
On the contrary, the display name is free to change as it wants.

radarsat1:
I think supporting every possible naming scheme under the sun is outside the scope of RtAudio.  It makes sense to have a single short name to refer to an API, but supporting many ways to refer to the same API should be left up to applications.  I would propose then just choosing a short, lower-case name for each API, such as "core", "winmm", etc.  This is how they are referred to in [configure.ac](https://github.com/thestk/rtaudio/blob/master/configure.ac#L41) for example.  Then, if programs want to replace "winmm" with "WinMM" they can do so but I don't see it useful to encode this in RtAudio.  For look-up, programs can pre lower-case their strings before passing them to RtAudio, then we only have to support lower case.

jpcima:
Right. So I used a lower case name as identifier, and a ordinary name as presentation to the user.
I clarify the comment strings and update unit tests consequently.

For the normal names, I don't find necessary to include the OS name as is done in `tests/audioprobe.cpp`. It's not a useful information to populate the selections of a GUI with, that is of any interest to users.

radarsat1:
I pushed a commit proposing what I meant regarding using a map.  Then the data is data and the lookup functions are exceedingly simple.  What do you think?

radarsat1:
ps yes I think providing a single display name is better than maintain a list of them, thanks.

jpcima:
It would be alright I think, if not for one little problem.
I don't mind to much if std::map is user as a storage container.

A problem exists, I think when you query a name from the C api.
When you modify the return such that it's no longer a std::string reference which is obtained from the static storage, a `c_str` returns a memory which is freed and creates a crash.
I think the best is to leave the return as const-reference.

radarsat1:
Yep, you are right I forgot about the C Api, I'll fix.

radarsat1:
I thought it'd be better to just return pointers the actual strings in the C API, and in doing this I realized it's better to get rid of the map in favour of an array.  Added some test code for the C API.

jpcima:
It's good. I've arranged the build in such way that the link would pass in Windows CI.
I wrote the option `RTAUDIO_EXPORT` in cmake builds as well, to produce also a clean library export.

radarsat1:
Ah, that "replace with something nicer" comment was just supposed to remind me to replace it with an initializer list when we go to C++11 ;)  But your change made me realize it's not necessary to have it at all.

jpcima:
Above commit makes C API behave in the same way as C++, and adds a unit test to check.

About more little details, I saw display names changed in transition sometime, and a number don't respect the official naming/spacing/casing of these technologies.

"Pulse" -> "PulseAudio"
"OpenSoundSystem" -> "Open Sound System"
"Jack" -> "JACK"
"CoreAudio" -> "Core Audio"

With that aside, I think I'm good.

radarsat1:
Not sure I see the point of explicitly checking if the API is compiled when translating the name, unless we also exclude those names from the table, but I was trying to keep the logic simple by just translating the name regardless of whether the API is compiled.  If we want to keep the strings out of the build for non-compiled apis it would be simpler to just pull NULLs in the table instead of adding a function.  I would argue for (1) putting NULLs in the table, or (2) always translating the names and taking the word "compiled" out, just call them `rtaudio_get_api_name`, `rtaudio_get_api_display_name`.

To be specific, for (1) i'm refering to leaving my simpler logic,

```
const char *rtaudio_compiled_api_name(rtaudio_api_t api) {
    if (api < 0 || api >= RTAUDIO_API_NUM)
        return NULL;
    return rtaudio_api_names[api][0];
}
```

and doing something like,
```
#ifdef __UNIX_JACK
  { "jack"        , "Jack" },
#else
  { NULL        , NULL },
#endif
```

in the table, which would give the same result and not require scanning the list of compiled apis every time.  I think it's slightly messier in the code though, so I would prefer (2) perhaps.

radarsat1:
Implementation details aside, @garyscavone any objection to merging this?

radarsat1:
Yeah I thought of arguing that it could be done outside RtAudio but then... it's kind of useful for applications to be able to easily list this info so I think it's a nice pair of functions to add just as a utility. Also could be useful e.g. for selecting an audio backend on the command-line for audioprobe.

jpcima:
I've changed the last commit to match the behavior described in (2).

The previous reason for checking API availability in name methods was because such checks were made in C++ RtAudio, while C would exhibit a different behavior.
This said, I'm fully in agreement with solution (2).
- I remove such checks from c++ API
- rename methods and update the comments accordingly

@garyscavone I would appreciate such convenience in the library, it would be a saver of repeating same boilerplate in programs over and again

PS. the code check is failing arbitrarily with such an error `Could not connect to mirror.jmu.edu:80 (134.126.36.183), connection timed out`

radarsat1:
Don't worry I think Travis seems to be (again) having some problems lately.. I'll take a look at this soon and merge it. Thanks!

PR #141, opened 2018-05-24
Fix ALSA CTL handle not being closed in the default case

Adjacence:
In the case where the default audio device is loaded in probeDeviceOpen, the ALSA CTL handle was not being closed. In my case (running Debian 9), every call to snd_ctl_open created a PulseAudio thread in the back-end which only exits upon calling snd_ctl_close. So, every time a stream was opened, a dangling thread would be created. This edit seems to have fixed the issue.

### DIFF
--- a/RtAudio.cpp
+++ b/RtAudio.cpp
@@ -7162,6 +7162,7 @@ bool RtApiAlsa :: probeDeviceOpen( unsigned int device, StreamMode mode, unsigne
       }
       nDevices++;
     }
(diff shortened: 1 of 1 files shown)


csmacdonald95:
It appears there is something wrong with two of the build jobs on Travis

csmacdonald95:
Ugh, weird build errors again. I won't close and reopen this time because that's probably not the appropriate way to re-run the checks.

PR #143, opened 2018-06-03
avoid cmake error: "install Library TARGETS given no DESTINATION!"

sonoro1234:
My platform is windows, my compiler mingw-w64.
I got this error on cmake <code>install Library TARGETS given no DESTINATION!</code> and after googling I arrived to this, which means reverting part of PR  #126 @claudiocabral

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -129,7 +129,9 @@ endif (BUILD_TESTING)
 configure_file("rtaudio.pc.in" "rtaudio.pc" @ONLY)
 
 install(TARGETS rtaudio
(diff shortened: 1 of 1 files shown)


sonoro1234:
https://stackoverflow.com/questions/22278381/cmake-add-library-followed-by-install-library-destination/22280842#22280842

claudiocabral:
All seems very good to me, I don't have access to a windows machine but this still compiles smoothly on linux.

sonoro1234:
@radarsat1 @garyscavone  can this be commited?

PR #147, opened 2018-06-10
Removed Unused Function

tay10r:
There is a function that appears in most of the derived API classes, called `getStreamLatency`. Since it is implemented in the base class, and it is non-virtual, I take it that it was not meant to show up in the derived classes. This PR removes the prototype definition from the derived API classes. I tested the build with ALSA, JACK, and Pulse.

### DIFF
--- a/RtAudio.h
+++ b/RtAudio.h
@@ -874,7 +874,6 @@ public:
   void startStream( void );
   void stopStream( void );
   void abortStream( void );
-  long getStreamLatency( void );
 
   // This function is intended for internal use only.  It must be
   // public because it is called by the internal callback handler,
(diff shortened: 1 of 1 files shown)


garyscavone:
I think the idea was that the function is only declared for the APIs in which it is supported. If it isn’t declared in a particular API, then it is not declared. But I guess that is just “window-dressing,” since if it is defined in the base class, all classes inherit it.

tay10r:
@garyscavone Perhaps a better solution would be to have a function like this in the base class.

```c++
class RtApi {
public:
  virtual bool canCalculateLatency() const noexcept { return true; }
};
```

Then override the function in the classes that don't support latency calculation.

```c++
class RtApiExample final : public RtApi {
public:
  virtual bool canCalculateLatency() const noexcept override { return false; }
};
```

radarsat1:
I think this is handled by the documentation that says

> If the API does not report latency, the return value will be zero.

Anyways, i agree that these functions are declared but not defined, and additionally inaccessible, so better remove them imho.  I'll merge.

PR #148, opened 2018-06-10
Appveyor CI

tay10r:
I would suggest using Appveyor to test compilation for Windows. Appveyor is free but it can be kind of awkward to setup for organizations. Once you get it going, though, it's basically the same as Travis CI. This would help contributors like myself find out if a set of commits break a Windows build.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,10 @@
+platform: x64
+configuration: Debug
+
+before_build:
+  - mkdir build
+  - cd build
+  - cmake ..
+
+build:
+  project: build\ALL_BUILD.vcxproj


radarsat1:
tbh i don't really feel like maintaining yet another CI config.  But this does look pretty minimal.
On the other hand I did set up Travis to test building the Windows code paths using MingW which seemed to work ok.

> This would help contributors like myself find out if a set of commits break a Windows build.

I finally actually have a Windows machine at my disposal, I haven't set it up for development yet but I think I'll be able to handle any issues moving forward..

tay10r:
@radarsat1 I have a Windows machine, too. I use it from time to time to test cross platform builds on Visual Studio. I would still recommend using AppVeyor as well as MinGW because MSVC can't always compile code that works on GCC.

I know, for example, that MinGW's support for c++11 threads has been spotty in the past.

radarsat1:
Now that Travis supports Windows I think that might be a better avenue.

PR #162, opened 2018-10-26
Add loopback support to WASAPI

MarcusTomlinson:
This change allows WASAPI users to capture "loopback" streams from render devices.


MatzeOGH:
I had to define an output stream parameters before getting it to work.
Is that intended behavior?

PR #163, opened 2018-10-29
fix go build due to enum casts

zserge:
It was not a good idea to cast enum array from C++ to C. Solved it with the simplest idea that came to mind. Maybe there are better ways, but this one worked, so I left it.


radarsat1:
Sorry, I merged this but then reverted it because I realized it completely breaks the build.  Did you test this?

zserge:
Hm... Tested with Go only. I'll have a look.

PR #169, opened 2018-12-05
rtaudio_c: add rtaudio_num_compiled_api()

sonoro1234:
In the c implementation: The const unsigned int rtaudio_num_compiled_apis is not exported in ffi interfaces of modules, so it is here returned by a function.

### DIFF
--- a/rtaudio_c.cpp
+++ b/rtaudio_c.cpp
@@ -23,7 +23,7 @@ const rtaudio_api_t *rtaudio_compiled_api() {
 }
 
 extern "C" const unsigned int rtaudio_num_compiled_apis;
-unsigned int rtaudio_num_compiled_api(void) {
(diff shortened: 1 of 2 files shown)


sonoro1234:
Just saw that this PR can be avoided if we look for RTAUDIO_API_UNSPECIFIED as the last entry always included.

radarsat1:
It's a bit confusing to have a function named `rtaudio_num_compiled_api` and a variable named `rtaudio_num_compiled_apis`.

Why not just use the `rtaudio_num_compiled_apis` variable directly?  It could be added `extern` to `rtaudio_c.h`.  Otherwise I suppose I would name it `rtaudio_get_num_compiled_apis`.

sonoro1234:
I agee, I will add new commit

PR #188, opened 2019-02-21
rename uninstall target to uninstall_rtaudio

sonoro1234:
when creating a cmake with GLFW and RtAudio as submodules, there is a cmake complain about unistall target already being defined in GLFW
My solution was to rename uninstall to uninstall_rtaudio

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -302,7 +302,7 @@ configure_file(
     "${CMAKE_CURRENT_BINARY_DIR}/RtAudioConfigUninstall.cmake" @ONLY)
 
 # Create uninstall target.
-add_custom_target(uninstall
+add_custom_target(uninstall_rtaudio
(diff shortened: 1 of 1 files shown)


sonoro1234:
Anyway, I worked a solution with ExternalProject_add which is much more complicated but is the way to go with third party libraries.

radarsat1:
Seems this was subsumed by #190 in any case.

sonoro1234:
> Seems this was subsumed by #190 in any case.

Not at all, instead of add_subdirectory (that I find to be a very simple option for using third party projects as submodules but can generate target name collisions) I used another method with ExternalProject_add in https://github.com/sonoro1234/anima

radarsat1:
Sorry my mistake, getting confused between projects.  I meant https://github.com/thestk/rtmidi/pull/190 for which I just made an analog commit in RtAudio https://github.com/thestk/rtaudio/commit/4ba65da5277a6b12b69e40877adaff03ae644ed1

PR #192, opened 2019-03-05
cmake add SKIP_HAVE_GETTIMEOFDAY

sonoro1234:
by defining SKIP_HAVE_GETTIMEOFDAY in cmake we can avoid use HAVE_GETTIMEOFDAY even if we have it

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,7 +53,7 @@ option(RTAUDIO_API_CORE "Build CoreAudio API" ${APPLE})
 # Check for functions
 include(CheckFunctionExists)
 check_function_exists(gettimeofday HAVE_GETTIMEOFDAY)
-if (HAVE_GETTIMEOFDAY)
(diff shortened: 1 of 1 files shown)


radarsat1:
I think you can just set it directly?

```
cmake .. -DHAVE_GETTIMEOFDAY=OFF
```

seems to work.  Curious about why you'd want to turn it off.

sonoro1234:
> I think you can just set it directly?
> 
> ```
> cmake .. -DHAVE_GETTIMEOFDAY=OFF
> ```
> seems to work.

I am working in bundled distribution and setting this affects to libsndfile library.

> Curious about why you'd want to turn it off.

In windows ASIO and WASAPI but not DS are giving streamTime values wildly variable (WASAPI even no monotonic)

sonoro1234:
I could set different values in cmake for rtaudio and libsndfile
Thanks

PR #193, opened 2019-03-26
Update CMake build system

frabert:
I am writing a portfile for [vcpkg](https://github.com/Microsoft/vcpkg) to be able to use RtAudio as a dependency.

As RtAudio already uses CMake, I thought I could try to update the build system to better use more modern CMake conventions and practices, which should allow a smoother integration with other projects.

Here is an overview of the changes:
- The project now only builds either shared _or_ static libraries
- The exported targets include the dependency on pthreads when it is built with it
- The Debug suffix for the Windows library has been removed, CMake automatically handles the build type with the `--config` flag

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 # Set minimum CMake required version for this project.
-cmake_minimum_required(VERSION 3.0 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.10 FATAL_ERROR)
 
 # Define a C++ project.
 project(RtAudio LANGUAGES CXX)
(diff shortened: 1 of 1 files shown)


radarsat1:
Okay from CMake docs I see that you can select static library using `BUILD_SHARED_LIBS`.  I would ask a review from users of the CMake build.

frabert:
Yes, that is the recommended way of specifying library linkage. Also if the "d" suffix is still needed for compatibility purposes maybe a default-on option can be added.

radarsat1:
I was hoping for some more feedback but this seems reasonable so I will merge it.

radarsat1:
I've just merged this only to realize it conflicts directly with changes in 4ba65da5277a6b12b69e40877adaff03ae644ed1 made to correspond with RtMidi #190: https://github.com/thestk/rtmidi/pull/190/commits/f517f47390d7297d82167518a82f385e1c385ea8

Specifically that commit changed BUILD_SHARED_LIBS etc to RTAUDIO_BUILD_SHARED_LIBS in order to "make use with add_subdirectory easier."  This commit changes RTAUDIO_BUILD_SHARED_LIBS to BUILD_SHARED_LIBS (etc.) in order to "update the build system to better use more modern CMake conventions and practices".

Care to comment please, @frabert, @mcclure

radarsat1:
I think the purpose there was that RTAUDIO_BUILD_SHARED_LIBS should have a namespace for the case that it is simply a subdirectory of another project and one wants to be able to have specific options for RtAudio/RtMidi that may differ from the main project.

I think a good compromise for now may be to use RTAUDIO_BUILD_SHARED_LIBS, but to set it to the value of BUILD_SHARED_LIBS if it is not already set.  I will do this for now.

frabert:
@radarsat1 Sorry, I wasn't aware of the issue with RtMidi. Your proposal seems like a fine compromise to me.

radarsat1:
It's okay, trying to cover many use cases at the same time is always challenging.  I'll see if @mcclure has any comments but I've done some ad-hoc testing and I think this compromise works.  I haven't specifically tested RtAudio or RtMidi in an add_subdirectory configuration where it has different build settings from the parent though.  Maybe I should do that, but busy updating other things at the moment, so if anyone wants to confirm for me that would be great.

PR #200, opened 2019-04-21
Reduce CMake required version to 3.5.

9define:
Recently the required CMake version was increased to 3.10. The CMake file in master does not require greater than version 3.0. This breaks compatibility with Ubuntu Xenial, Raspbian Stretch, and presumably many other GNU/Linux distros where the upstream CMake version that's available is greater than 3.0 but much less than 3.10. I have chosen 3.5 for this PR as it's what's required by Xenial, but in general the required version should be reduced. I have at least one project is broken because of this change in the most recent release. If this PR is merged I would ask for a small amendment release for those folks that rely on stable releases. Thanks for your consideration.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 # Set minimum CMake required version for this project.
-cmake_minimum_required(VERSION 3.10 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.5 FATAL_ERROR)
 
 # Define a C++ project.
 project(RtAudio LANGUAGES CXX)


radarsat1:
My inclination is to avoid regressing things.  Can you describe your setup and how this broke it?  Perhaps there is another solution.  Xenial is 3 years old after all.  I'd like if you could justify why simply patching on your end is not a sufficient solution.

9define:
@radarsat1 Sorry for the delayed response, I saw your reply and forgot to get back to this thread.

I understand the desire not to regress changes, but my overall reason for wanting this change (or to return to the previously required version of 3.0) is that this project does not require anything newer than 3.0 to be configured/built properly, as far as I can tell. I generally only bump my required CMake version for a project, if I start using a feature that is only available in a newer version or there are important bug fixes related to the features I use. Of course, that is my style/workflow so this project may have another philosophy (if so, please let me know).

A bit about my setup. I have a project in development (so it's not yet public) that uses git submodules to grab build-time dependencies, such as this project. I also have some CI runners for distros/releases I wish to target for future release of my project. Those distros are Xenial, Bionic, Raspbian/Debian Stretch, Fedora 30, and Arch. Obviously, Xenial is getting older, but I think that enough people/companies are still using it that releasing for it should be supported, as long as it doesn't negatively affect the quality of my codebase.

Not much in my defense, but it's not terribly difficult to patch this myself, but I have to do so on my own branch of my own fork of this project, which I (and presumably others) would rather not do.

Anyways, my general point is that I don't see a reason to bump the required version without wanting new features/needing bug fixes, but not supporting Xenial is something to consider for my project. Thanks for your consideration.

david-c14:
I've just found this page precisely because I couldn't build rtaudio on Ubuntu 16.04 
The latest version of cmake in the Ubuntu distribution is 3.5.1
If cmake 10 is not really necessary for building rtaudio, it's a considerable extra burden to have to deal with

garyscavone:
@radarsat1 Can you deal with this? I have no experience with cmake. It seems like a relatively minor change to me but you expressed some concern above.

radarsat1:
Well, it is a small change to make and we can certainly revert the required CMake version, but should we in general take on the burden of supporting 4-year-old Linux distributions?  I feel that if people want to use the latest version of a dependency while targetting old systems, the burden to backport things is on them.

Anyways, at the time I didn't want to put out a new release just for this, but maybe we can include the reversion in the next release.

PR #207, opened 2019-06-19
Change type of captureFlags to DWORD

mcwissink:
When compiling with WASAPI on my windows machine, the build fails since the type of a variable captureFlags (unsigned long) does not match the windows DWORD type (unsigned int) in a function call. Changing the captureFlags type to DWORD seems to be the most reliable fix for this issue since it ensures that captureFlags is processor independent.


MarcusTomlinson:
Well spotted @mcwissink, thanks!

MarcusTomlinson:
@radarsat1 or @garyscavone, I’m happy with this change, could you ack and merge?

Not a blocker, but I am wondering why CI didn’t chime in on this PR. Any idea?

PR #213, opened 2019-09-04
PulseAudio: get preferredSampleRate

sreimers:
This is a basic proof of concept to get the really preferred sample rate from pulseaudio async api.

### DIFF
--- a/RtAudio.cpp
+++ b/RtAudio.cpp
@@ -8482,7 +8482,7 @@ static void rt_pa_server_callback(pa_context *context, const pa_server_info *inf
   (void)data;
   pa_sample_spec ss;
 
-  if (info == nullptr)
+  if (!info)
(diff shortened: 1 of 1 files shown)


garyscavone:
I don't have an easy way to test this. You say a "proof of concept". Is it something that should be merged?

sreimers:
i've been using the code in my studio link project for several months now and haven't had any user complaints, so i think it should be safe to merge it. its the only way to get the real samplerate so i can resample if necessary.

radarsat1:
I'll test it

radarsat1:
Yes, it seems to work reliably, and I don't see a cleaner way to do this, so I'll merge.  Leaving the default 48000 for now too since I don't see any examples in other APIs where it can be left as 0 so I assume user code assumes it is non-zero.

Issue #214, opened 2019-09-20
Mingw application crashed when using wasapi resampler and random malloc/free activity


Serg77m:
I have problems using wasapi for mingw under Windows 7. The application crashes if a non-standard samplerate is used and at the same time there is some activity malloc()/free(). Tested on different versions of mingw. Compilation flags:

g++ main.cpp RtAudio.cpp -D__WINDOWS_WASAPI__ -lmfplat -lmfuuid -lksguid -lwmcodecdspuuid -lole32 -static-libgcc -static-libstdc++ -mthreads

Code:

#include "RtAudio.h"
#include <iostream>
#include <cstdlib>
#include <cstring>
// Pass-through function.
int inout( void *outputBuffer, void *inputBuffer, unsigned int nBufferFrames,
           double streamTime, RtAudioStreamStatus status, void *data )
{
  if ( status ) std::cout << "Stream over/underflow detected." << std::endl;
  unsigned long *bytes = (unsigned long *) data;
  memcpy( outputBuffer, inputBuffer, *bytes );
  return 0;
}
int main()
{
 RtAudio adac;
  if ( adac.getDeviceCount() < 1 ) {
    std::cout << "\nNo audio devices found!\n";
    exit( 0 );
  }
  // Set the same number of channels for both input and output.
  unsigned int bufferBytes, bufferFrames = 512;
  RtAudio::StreamParameters iParams, oParams;
  iParams.deviceId = adac.getDefaultInputDevice();
  iParams.nChannels = 2;
  oParams.deviceId = adac.getDefaultOutputDevice();
  oParams.nChannels = 2;
  try {
    adac.openStream( &oParams, &iParams, RTAUDIO_SINT32, 46000, &bufferFrames, &inout, (void *)&bufferBytes );
  }
  catch ( RtAudioError& e ) {
    e.printMessage();
    exit( 0 );
  }
  bufferBytes = bufferFrames * 2 * 4;
  try {
    adac.startStream();
  }
  catch ( RtAudioError& e ) {
    e.printMessage();
    goto cleanup;
  }

  // Random malloc/free activity
  void *randomData[1024];
  memset(randomData, 0, sizeof(randomData));
  for(;;) {
    unsigned int blockNo = rand() & 1023;
    if (randomData[blockNo] != NULL)
      free(randomData[blockNo]);
    unsigned int blockSize = (rand() & 32767) + 8;
    randomData[blockNo] = malloc(blockSize);
  }

  adac.stopStream();
[...]

PR #215, opened 2019-09-22
Update RtAudio.cpp

Serg77m:
I found solution for issue #214


MarcusTomlinson:
Thank you for taking the time to look into this! It’s very much appreciated. I’ll try to review sometime this week.

Do you perhaps know where the access violation is occurring and/or why this change fixes it?

Serg77m:
I don’t know exactly why, but in the process of resampling, sometimes one extra sample comes out. One came in, two came out. Maybe, this is due to interpolation. And it does not fit into the buffer.

PR #230, opened 2020-02-11
Memory Leak using Alsa on Linux ( RtApiAlsa::probeDeviceOpen )

### LINKED ISSUE #228: Memory Leak using Alsa on Linux ( RtApiAlsa::probeDeviceOpen )
### RtAudio.cpp

in the RtApiAlsa::probeDeviceOpen (7529)

in the debug you attach the output to stderr

```c++
#if defined(__RTAUDIO_DEBUG__)
	snd_output_t *out;
	snd_output_stdio_attach(&out, stderr, 0);
#endif
```

But you don't release the output on the end of the method... each time you leak 16 to 24 bytes due to this attachment.

### Solution

 - before each return in this function put something like this:

```c++
#if defined(__RTAUDIO_DEBUG__)
//stops mem leak at debug
snd_output_close(out);
#endif
return FAILURE;
```

And finally you can return success with no leaks.

```c++
#if defined(__RTAUDIO_DEBUG__)
//stops mem leak at debug
snd_output_close(out);
#endif
return SUCCESS;
```

That's it.

A-Ribeiro:
Closes #228

### DIFF
--- a/RtAudio.cpp
+++ b/RtAudio.cpp
@@ -7542,6 +7542,10 @@ bool RtApiAlsa :: probeDeviceOpen( unsigned int device, StreamMode mode, unsigne
       if ( result < 0 ) {
         errorStream_ << "RtApiAlsa::probeDeviceOpen: control open, card = " << card << ", " << snd_strerror( result ) << ".";
         errorText_ = errorStream_.str();
+
+#if defined(__RTAUDIO_DEBUG__)
+        snd_output_close(out);
+#endif
         return FAILURE;
       }
       subdevice = -1;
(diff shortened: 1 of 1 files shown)


garyscavone:
Sorry it took so long to look at this. The proposed solution is rather repetitive. What about changing every "return FAILURE;" to "goto error;" and then add the check for DEBUG there? You'd also have to do a check before the "return SUCCESS;".

I don't have a Linux box handy to test this.

A-Ribeiro:
Hi, 

i personally, don't like to use goto with cpp. 

but I agree that will be less code with this. 

i detected the leak using the libasan with sanitize address flag... the leak appeared in that point.

radarsat1:
tested, thanks

radarsat1:
By the way I had a brief look at the `goto` solution (which I like) but there are many instances of goto crossing initialization errors, e.g.

```
RtAudio.cpp:7881:15: note:   crosses initialization of ‘AlsaHandle* apiInfo’
   AlsaHandle *apiInfo = 0;
```

Then I came up with an RAII solution which is better, will push it now.

PR #241, opened 2020-04-10
AVX optimizations for the convertBuffer function

eloi-loomai:


garyscavone:
Why was this closed?

radarsat1:
This is great, only it makes me wish we had unit tests!
@eloi-loomai have you tried to see if autovectorization can get to the same result?  I find it much cleaner and less error prone than manual intrinsics when you can get it to work.  Have you done performance testing?

garyscavone:
I agree that these changes would make the code harder to maintain and update in the future. I'm inclined to close this PR without merging it, unless someone can convince me otherwise in the next few days.

Issue #248, opened 2020-05-08
WASAPI Input stream issue


jkeller51:
Hi,

Up until recently I couldn't use WASAPI at all or I'd get a crash. But with the latest version of RtAudio a1ddca0c4, I was excited to find that I can use it now.

However, if I use RtAudio in duplex mode with microphone input selected and using WASAPI driver, I get a weird glitchy-sounding microphone signal overlayed in my output, even when I just fill the output with 0s, e.g.:

```c++
static int RtAudioStreamCallback(void * outputBuffer, void * inputBuffer, unsigned int nBufferFrames, double streamTime, RtAudioStreamStatus status, void * data) {
    /* data should be a pointer to a ZivixPlugin */
    ZivixPlugin * zivixPlugin = (ZivixPlugin *) data;
    uint32_t och = zivixPlugin->getNumOutputChannels();
    uint32_t ich = zivixPlugin->getNumInputChannels();
    double * out = (double *) outputBuffer;
    for (uint32_t ch=0; ch<och; ch++) {  // non-interleaved mode
        for (uint32_t s=0;s<nBufferFrames;s++) {
            out[ch*nBufferFrames + s] = 0;
        }
    }
    return 0;
}
```

This callback code causes weird distorted microphone input coming through to my output.

I've been using DirectSound and ASIO without any audio issues.

I discovered that if I set the input settings to NULL (so we are in playback-only mode), the issue goes away.

MarcusTomlinson:
Could you please run RtAudio's own duplex test app locally and see if the problem persists:
https://github.com/thestk/rtaudio/blob/master/tests/duplex.cpp

I've just given this ^ a go locally without issue.

jkeller51:
I ran the test with WASAPI API and FLOAT64 data type. I had the same type of audio issues. Github doesn't allow embedding videos, so here is a link:

https://streamable.com/9ty1ra

jkeller51:
I changed the audio callback to clear the output buffer

```c++
int inout( void *outputBuffer, void *inputBuffer, unsigned int /*nBufferFrames*/,
           double /*streamTime*/, RtAudioStreamStatus status, void *data )
{
  // Since the number of input and output channels is equal, we can do
  // a simple buffer copy operation here.
  if ( status ) std::cout << "Stream over/underflow detected." << std::endl;

  unsigned int *bytes = (unsigned int *) data;
  //memcpy( outputBuffer, inputBuffer, *bytes );
[...]

PR #249, opened 2020-05-10
convertBuffer(): clear duplex device outBuffer if outJump > inJump

MarcusTomlinson:
Resolves: https://github.com/thestk/rtaudio/issues/248

### DIFF
--- a/RtAudio.cpp
+++ b/RtAudio.cpp
@@ -10390,9 +10390,8 @@ void RtApi :: convertBuffer( char *outBuffer, char *inBuffer, ConvertInfo &info 
   // data interleaving/deinterleaving.  24-bit integers are assumed to occupy
   // the lower three bytes of a 32-bit integer.
 
-  // Clear our device buffer when in/out duplex device channels are different
-  if ( outBuffer == stream_.deviceBuffer && stream_.mode == DUPLEX &&
-       ( stream_.nDeviceChannels[0] != stream_.nDeviceChannels[1] ) )
+  // Clear our duplex device output buffer if there are more device outputs than user outputs
+  if ( outBuffer == stream_.deviceBuffer && stream_.mode == DUPLEX && info.outJump > info.inJump )
     memset( outBuffer, 0, stream_.bufferSize * info.outJump * formatBytes( info.outFormat ) );
 
   int j;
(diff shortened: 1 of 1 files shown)


MarcusTomlinson:
Hmm, so I get that the < was put here as an optimization, but I’m actually wondering if the original intention was >.

I.e. if outputs > inputs, clear the buffer.

This to me seems logical. It doesn’t matter if the device buffer has some output data still floating around in it when converting for user input as the user will only ever receive the channels he/she cares about.

As it stands now (if inputs > outputs), input data left behind in the device buffer after convertBuffer will always end up in the output.

MarcusTomlinson:
Hmm, so I get that the < was put here as an optimization, but I’m actually wondering if the original intention was >.

I.e. if outputs > inputs, clear the buffer.

This to me seems logical. It doesn’t matter if the device buffer has some output data still floating around in it when converting for user input as the user will only ever receive the channels he/she cares about.

As it stands now (if outputs < inputs), input data left behind in the device buffer after convertBuffer will always end up in the output.

@radarsat1 & @garyscavone, does this logic make sense? Or can you recall a particular scenario where clearing the device buffer is needed for outputs < inputs?

MarcusTomlinson:
Just pushed an even simpler, optimised solution to this:

**Clear the duplex device's output buffer if there are more device outputs than user outputs.**

Note that outJump will never be greater than inJump in the reverse direction (from device input to user input), as the user can't ask for more input channels than are physically available.

radarsat1:
I'll leave this one for @garyscavone as i'm not very familiar with this part of the code.

MarcusTomlinson:
@garyscavone, have you managed to have a look at this at all?

garyscavone:
Your logic seems correct. I assume you have tested the change and it worked, so go ahead and commit it. If you can't do the commit for some reason, let me know and I will.

MarcusTomlinson:
> Your logic seems correct. I assume you have tested the change and it worked, so go ahead and commit it. If you can't do the commit for some reason, let me know and I will.

Yeah, I've tested it. It certainly fixes the linked bug, and I've not seen any further regression caused. Thanks for the thumbs up, I'll merge.

PR #258, opened 2020-06-04
Add override keyword

radarsat1:
This adds the "override" keyword which C++11 compilers warn against forgetting, causing problems with `-Werror`.


radarsat1:
This seems to be needed more as I work on other branches so I'm going to go ahead and merge it.

PR #261, opened 2020-07-01
Add wasapi shared mode period

shangjiaxuan:
Wasapi on win10 desktop and later can query the system shared mode mixer's period for less latency and less overhead by adjusting buffer size to accomodate it.

### DIFF
--- a/RtAudio.cpp
+++ b/RtAudio.cpp
@@ -3799,6 +3799,86 @@ if ( objectPtr )\
 
 typedef HANDLE ( __stdcall *TAvSetMmThreadCharacteristicsPtr )( LPCWSTR TaskName, LPDWORD TaskIndex );
 
+// Copied from header
+#ifndef __IAudioClient2_INTERFACE_DEFINED__
+#define __IAudioClient2_INTERFACE_DEFINED__
+
+/* interface IAudioClient2 */
+/* [local][unique][uuid][object] */
+
+EXTERN_C const IID IID_IAudioClient2;
+
+MIDL_INTERFACE("726778CD-F60A-4eda-82DE-E47610CD78AA")
+IAudioClient2 : public IAudioClient
+{
+public:
+    virtual HRESULT STDMETHODCALLTYPE IsOffloadCapable(
+        /* [annotation][in] */
+        _In_  AUDIO_STREAM_CATEGORY Category,
+        /* [annotation][out] */
+        _Out_  BOOL * pbOffloadCapable) = 0;
+
+    virtual HRESULT STDMETHODCALLTYPE SetClientProperties(
+        /* [annotation][in] */
+        _In_  const AudioClientProperties* pProperties) = 0;
+
+    virtual HRESULT STDMETHODCALLTYPE GetBufferSizeLimits(
+        /* [annotation][in] */
+        _In_  const WAVEFORMATEX* pFormat,
+        /* [annotation][in] */
+        _In_  BOOL bEventDriven,
+        /* [annotation][out] */
+        _Out_  REFERENCE_TIME* phnsMinBufferDuration,
+        /* [annotation][out] */
+        _Out_  REFERENCE_TIME* phnsMaxBufferDuration) = 0;
+
+};
+#endif
+
+#ifndef __IAudioClient3_INTERFACE_DEFINED__
+#define __IAudioClient3_INTERFACE_DEFINED__
+
+/* interface IAudioClient3 */
+/* [local][unique][uuid][object] */
+
+
+EXTERN_C const IID IID_IAudioClient3;
+
+MIDL_INTERFACE("7ED4EE07-8E67-4CD4-8C1A-2B7A5987AD42")
+IAudioClient3 : public IAudioClient2
+{
+public:
+    virtual HRESULT STDMETHODCALLTYPE GetSharedModeEnginePeriod(
+        /* [annotation][in] */
+        _In_  const WAVEFORMATEX * pFormat,
+        /* [annotation][out] */
+        _Out_  UINT32 * pDefaultPeriodInFrames,
+        /* [annotation][out] */
+        _Out_  UINT32 * pFundamentalPeriodInFrames,
+        /* [annotation][out] */
+        _Out_  UINT32 * pMinPeriodInFrames,
+        /* [annotation][out] */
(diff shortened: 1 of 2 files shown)


shangjiaxuan:
Windows SDK of earlier versions may not have this function exported. Thus a runtime load of the IID (define the GUID in the sorce file, and call Activate there) with a forwad declaration of the interface may be better, and that would require a conditional on the sdk version, and define the interface if not existing.

RtAudio.cpp
```diff
@@ -47,6 +47,7 @@
 #include <cstring>
 #include <climits>
 #include <cmath>
+#include <cstring>
```
shangjiaxuan, line 50: This is not necessary since the hack in the first commit is no longer present and the openStream function always have a updated buffersize already.

RtAudio.h
```diff
@@ -1064,6 +1064,8 @@ class RtApiWasapi : public RtApi
   static DWORD WINAPI stopWasapiThread( void* wasapiPtr );
   static DWORD WINAPI abortWasapiThread( void* wasapiPtr );
   void wasapiThread();
+
+  static bool iaudioclient3_available;
```
shangjiaxuan, line 1068: adding static variable here will not affect abi since it's never allocated on the stack.

RtAudio.cpp
MarcusTomlinson: If it's not necessary, why is it still here?

RtAudio.cpp
```diff
@@ -5301,26 +5370,35 @@ void RtApiWasapi::wasapiThread()
...
+          renderResampler->Convert(convBuffer,
+                                    stream_.deviceBuffer,
+                                    stream_.bufferSize,
+                                    convBufferSize);
+          // Push callback buffer into outputBuffer
+          callbackPushed = renderBuffer.pushBuffer(convBuffer,
+                                                    convBufferSize * stream_.nDeviceChannels[OUTPUT],
+                                                    stream_.deviceFormat[OUTPUT]);
+        }
+        else if (renderResampler->get_sampleratio()==1) {
```
MarcusTomlinson, line 5383: This is not necessary as this check is done in renderResampler->Convert()

MarcusTomlinson:
Thanks for this @shangjiaxuan, I'll try to have a closer look over the next few weeks.

RtAudio.cpp
shangjiaxuan: The second commit is not intended, the second commit was trying to reduce a few copies when there's no need. I'm really sorry for this messy setup of pr.

MarcusTomlinson: No worries, could you please made the necessary pushes to your branch such that only the required changes remain. Thanks!

shangjiaxuan:
Finished cleaning up, I'm really sorry for the inconvenience.

RtAudio.cpp
```diff
@@ -4796,14 +4890,24 @@ bool RtApiWasapi::probeDeviceOpen( unsigned int device, StreamMode mode, unsigne
...
+    }
     if ( FAILED( hr ) ) {
       errorText_ = "RtApiWasapi::probeDeviceOpen: Unable to retrieve render device audio client.";
       goto Exit;
     }
 
-    hr = renderAudioClient->GetMixFormat( &deviceFormat );
+    if(iaudioclient3_available){
+        IAudioClient3* client3 = (IAudioClient3*)renderAudioClient;
+        hr = client3->GetCurrentSharedModeEnginePeriod(&deviceFormat,bufferSize);
```
shangjiaxuan, line 4907: This will make the OpenStream function's parameter updated to be the system mixer buffer size. Perhaps a flag somewhere to disable the negotiation and always respect the user's input can be used.

garyscavone:
@MarcusTomlinson ... where do things stand with this PR?

MarcusTomlinson:
> @MarcusTomlinson ... where do things stand with this PR?

I don't know. I'm not particularly happy about this: https://github.com/thestk/rtaudio/pull/261/files#diff-7f8796d083d3c648a4b988425f73d69fce7dafad0a9a617c2c80899eb18bb07bR3802

In my opinion it's nice to have but not important enough to warrant muddying up the code this much.

MarcusTomlinson:
@shangjiaxuan are you able to motivate with real measurements how effective this change would be?

erwincoumans:
What is the minimum latency/buffer size you can get? Are there driver settings that can influence this?
I tried RtAudio with __WINDOWS_WASAPI__ (with and without this patch) and get stuck at a latency of 10ms, a minimum buffer size of 480 on Windows @48000Hz, while on MacOS (CoreAudio) latency/buffer size can be much lower.
~~~
 RtAudioFormat format = RTAUDIO_FLOAT32;
    RtAudio::StreamParameters parameters;
    parameters.deviceId = audioDevice >= 0 ? audioDevice : dac.getDefaultOutputDevice();
    parameters.nChannels = 2;
    bool dacOK = true;

    unsigned int bufferFrames = 480;
    try {
        dac.openStream(&parameters,NULL,format,(unsigned int)sampleRate,&bufferFrames,&tick,(void*)&data1);
    }
    catch(RtAudioError& error) {
        error.printMessage();
        dacOK = false;
    }
~~~

gillesdegottex:
I faced the same issue @erwincoumans 

Based on my understanding, this is due to Windows audio design limitation:
https://docs.microsoft.com/en-us/windows-hardware/drivers/audio/low-latency-audio

Namely (for both render and capture):
> Before Windows 10, this buffer was always set to ~10ms.

Then (for both render and capture):
* Starting with Windows 10, the buffer size is defined by the audio driver

So if you make a capture/render system (like RtAudio), you were bound to 10ms limitation.
And from my experiment, I've the impression that this limitation/default value still holds somewhere (by the audio driver?).

On both Linux and MacOS I can receive perfectly regular 192 samples buffers at 48kHz, each 4ms exactly (4*48 -> 4ms).
On Windows I do receive 192 samples buffers, but the intervals are more like 10ms, then some at 0ms intervals, and so on (I don't remember the exact pattern, though not regular at all).

MarcusTomlinson:
I'm so sorry this took me so long to get to @shangjiaxuan!

Thank you so much for this PR, it helped a lot in implementing https://github.com/thestk/rtaudio/commit/f115d797dd65ddbc5d2fdd2a568c51e8a4d768ed.

Closing now as fixed.

PR #264, opened 2020-07-19
RFC: ASIO + Wine build fixes

falkTX:
The changes in this PR allows to build ASIO with Wine, which is needed for [WineASIO](https://github.com/wineasio/wineasio/) project.

I understand it is a weird situation, so I am even wondering if you would consider such changes in the first place...
Comments welcome.

### DIFF
--- a/include/asio.cpp
+++ b/include/asio.cpp
@@ -10,6 +10,9 @@
 	
 #include <string.h>
 #include "asiosys.h"		// platform definition
+#if WINDOWS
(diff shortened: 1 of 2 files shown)


garyscavone:
We are already "out on a limb" by redistributing the asio files, so I do not feel comfortable making changes to those. If someone wants to compile with Wine, they can just do a search and replace on their own copy of the asio file.

PR #269, opened 2020-08-08
add userData in the error callback

gillesdegottex:
Might be useful for contextualise an error.


garyscavone:
I'm closing this for the same reasons as stated in PR #267 ... I think we should be moving toward the behaviour in the noexceptions branch.

gillesdegottex:
I don't quite see how errors raised while the stream is open should be dealt with in the noexceptions branch.
Do you have a few elements about that?

Issue #265, opened 2020-07-29
Memory error at stopStream/closeStream


TpnyF:
There are problems with stopStream/closeStream - it's really a rule.

I tried the versions: 4.1.2 , 5.0.0 , 5.1.0 They all give memory error when I close or stop the stream. Is there a problem with my code?

If give an error to such a simple application, and even on your example, I wonder how it behaves in complex applications.

`````
#include <windows.h>
#include "RtAudio.h"
#include <iostream>
#include <cstdlib>

enum Menu {
	id_play = 0,
	id_stop
};
typedef signed short MY_TYPE;
#define FORMAT RTAUDIO_SINT16
#define SCALE  32767.0

#define SLEEP( milliseconds ) Sleep( (DWORD) milliseconds ) 
#define BASE_RATE 0.005
#define TIME   1.0

HINSTANCE mHInstance;

void errorCallback( RtAudioError::Type type, const std::string &errorText )
{
  // This example error handling function does exactly the same thing
  // as the embedded RtAudio::error() function.
  std::cout << "in errorCallback" << std::endl;
  if ( type == RtAudioError::WARNING )
    std::cerr << '\n' << errorText << "\n\n";
  else if ( type != RtAudioError::WARNING )
    throw( RtAudioError( errorText, type ) );
}

unsigned int channels;
RtAudio::StreamOptions options;
unsigned int frameCounter = 0;
bool checkCount = false;
unsigned int nFrames = 0;
const unsigned int callbackReturnValue = 1;
unsigned int bufferFrames, fs, device = 0, offset = 0;
double *data = (double *) calloc( channels, sizeof( double ) );


RtAudio dac;

int saw( void *outputBuffer, void * /*inputBuffer*/, unsigned int nBufferFrames,
         double /*streamTime*/, RtAudioStreamStatus status, void *data )
{
  unsigned int i, j;
  extern unsigned int channels;
  MY_TYPE *buffer = (MY_TYPE *) outputBuffer;
  double *lastValues = (double *) data;

  if ( status )
    std::cout << "Stream underflow detected!" << std::endl;

  double increment;
  for ( j=0; j<channels; j++ ) {
    increment = BASE_RATE * (j+1+(j*0.1));
    for ( i=0; i<nBufferFrames; i++ ) {
      *buffer++ = (MY_TYPE) (lastValues[j] * SCALE * 0.08);
      lastValues[j] += increment;
[...]

PR #278, opened 2020-12-03
Issue 265 Memory error at stopStream/closeStream

desmondlr:
Addressing issue #265.

I think error was produced by accessing recently freed memory due to race conditions in RtApiWasapi. I've added mutual exclusion to several functions in RtApiWasapi, as well as RtApi::error() for thread safety.


garyscavone:
I would generally not approve this PR because I do not want to increase the use of mutexes, especially in the RtApi class. In fact, I would suggest we try to get away from the mutexes as much as possible. Since the problem this PR is attempting to create seems focused on the WASAPI API, changes should be focused on that implementation and not make their way out to the general RtApi functions. Plus, there are various checks that are failing in Travis. I generally defer to @MarcusTomlinson regarding WASAPI but please no changes to RtApi to fix a WASAPI problem.

PR #279, opened 2021-01-17
Fix convertBuffer() when converting the value "1.0" from float32 or float64 to SINT32

AndrewBelt:
I'm dumb and used `long` instead of `long long` in a PR several months ago. On Linux/Mac this works fine, but on MSVC and Mingw, `sizeof(long) == 4` so `lround(1.0 * 2147483648.f)` was overflowing to 0x80000000.
Fixed by using `llround()` and `long long` literals.

### DIFF
--- a/RtAudio.cpp
+++ b/RtAudio.cpp
@@ -10584,7 +10584,8 @@ void RtApi :: convertBuffer( char *outBuffer, char *inBuffer, ConvertInfo &info 
       Float32 *in = (Float32 *)inBuffer;
       for (unsigned int i=0; i<stream_.bufferSize; i++) {
         for (j=0; j<info.channels; j++) {
-          out[info.outOffset[j]] = (Int32) std::min(std::lround(in[info.inOffset[j]] * 2147483648.f), 2147483647L);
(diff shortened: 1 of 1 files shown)


garyscavone:
Does this PR require some update to travis?

sonoro1234:
Travis is complaining about doxygen not installed in macosx build?!

MarcusTomlinson:
I say just remove this line:
https://github.com/thestk/rtaudio/blob/34a3752e0c8249dc1780d196cd24e745425f0c77/.travis.yml#L63

It’s not necessary.

AndrewBelt:
Looks like the Mac Travis build breakage is because brew no longer supports shallow checkouts. The issue is unrelated to this PR, but if someone wants to fix it, just check out brew without shallow.

garyscavone:
Can you provide further details on how to fix the Travis problem? I'm regretting agreeing to add the Travis support, since I have no experience with it, nor do I have the time to learn it.

garyscavone:
And why is Travis trying to install Doxygen? I don't see the need to be checking the documentation for problems. If someone knows how to remove that, please do.

PR #288, opened 2021-05-02
asiolist.cpp: remove unused variable

autoantwort:
That seems to cause MinGW build failures in debug builds, see microsoft/vcpkg#17617


garyscavone:
I generally would avoid modifying the ASIO files but this one seems relatively harmless. At the same time, I don't see why a compiler should require unused variables to be cleaned up.

dyfer:
> I don't see why a compiler should require unused variables to be cleaned up.

I think that's due to `-Werror` [here](https://github.com/thestk/rtaudio/blob/master/CMakeLists.txt#L76), isn't it?

PR #292, opened 2021-05-12
Fix WASAPI backend loading AVRT.dll with UNICODE defined

### LINKED ISSUE #291: WASAPI backend attempting to load "AVRT.dll" fails with UNICODE defined
https://github.com/thestk/rtaudio/issues/10#issuecomment-41206750

https://github.com/thestk/rtaudio/blob/8f8250e311946b6b831357bdd6a030db3b75c057/RtAudio.cpp#L4998

You can't cast a char-based string to a `LPCTSTR` (which if `UNICODE` is defined, is a `wchar_t const*`), and pass it to LoadLibrary (which gets defined as `LoadLibraryW` and expects UTF-16 input). Currently, RtAudio tries loading `噁呒搮汬.DLL` instead.

Since the string you pass in is ASCII, the easiest solution is to call `LoadLibraryA` instead, but I'm not sure if it's better to call that, or to call `LoadLibraryW` explicitly and pass in a wide string.

nyanpasu64:
Previously, RtAudio casted a char-based string literal to a `TCHAR const*`, and passed it into `LoadLibrary` (a `#define`). With UNICODE defined, `TCHAR` is `wchar_t` and `LoadLibrary` maps to `LoadLibraryW` (which expects UTF-16 input). Windows interprets the string literal as UTF-16 and tries loading `噁呒搮汬.DLL` instead. Even worse, this performs an out-of-bounds read because when loading UTF-16 null-terminated strings, Windows looks for two consecutive null bytes. If the string literal's null terminator isn't followed by a second null byte in the program's memory layout, Windows might read even more gibberish and try to interpret it as UTF-16. (This is *probably* harmless but still bad.)

This PR changes it to unconditionally call `LoadLibraryW` and pass in a UTF-16 string literal. My guess is that this is slightly more efficient than calling `LoadLibraryA` (which allocates a new UTF-16 string, calls a locale-to-UTF-16 conversion routine, and passes the UTF-16 string to `LoadLibraryW`).

(Previously searching for AvrtDll would always fail on UNICODE builds. Now that it succeeds, I'm not certain that this won't expose misbehavior in the `if ( AvrtDll )` branch. But hopefully the branch works fine, and hopefully it's been sufficiently tested in non-UNICODE builds. I tried it in my app, and it didn't crash or anything.)

Fixes #291.

### DIFF
--- a/RtAudio.cpp
+++ b/RtAudio.cpp
@@ -4995,7 +4995,7 @@ void RtApiWasapi::wasapiThread()
   RtAudioError::Type errorType = RtAudioError::DRIVER_ERROR;
 
   // Attempt to assign "Pro Audio" characteristic to thread
-  HMODULE AvrtDll = LoadLibrary( (LPCTSTR) "AVRT.dll" );
+  HMODULE AvrtDll = LoadLibraryW( L"AVRT.dll" );
   if ( AvrtDll ) {
     DWORD taskIndex = 0;
     TAvSetMmThreadCharacteristicsPtr AvSetMmThreadCharacteristicsPtr =


MarcusTomlinson:
LGTM! And thanks for the detailed description!

Issue #270, opened 2020-09-09
ALSA: Missing "default" device


gillesdegottex:
The default device on ALSA is sometimes missing, or being busy, unaccessible from ALSA.
Eg. there is currently a known issue reported as `ALSA lib pulse.c:242:(pulse_connect) PulseAudio: Unable to connect: Connection terminated`

So the device id `0` is not always the default ALSA device.

In such situation, line here:
https://github.com/thestk/rtaudio/blob/d7f12763c55795ef8a71a9b589b39e7be01db7b2/RtAudio.cpp#L7562
just shifts all ALSA device ids, so that the wrong device id N+1 is picked up when opening the device id N.

In the same way, the sections

>   result = snd_ctl_open( &chandle, "default", SND_CTL_NONBLOCK );
  if ( result == 0 ) {
    if ( nDevices++ == device ) {
      strcpy( name, "default" );
      goto foundDevice;
    }
  }

in `getDeviceInfo` and `getDeviceCount` will always be sloppy as long as this "default" device appears or not.

As a potential solution, I suggest to not make special cases for the "default" device, just drop those sections and thus treat the "default" device as any other one, in order to be independent of its presence.

In order to implement RTAUDIO_ALSA_USE_DEFAULT in `probeDeviceOpen`, we could first search for the id of the device named "default" and then run the rest of `probeDeviceOpen` as for any other device (and fail if this id can't be found).

I'm happy to make PR if the analysis above looks correct to you and you like the solution.

Cheers,

radarsat1:
Related #224

gillesdegottex:
oh my, sry to have missed that!

antonrunov:
I hit very similar issue with
```ALSA lib pulse.c:243:(pulse_connect) PulseAudio: Unable to connect: Connection terminated```
error. But at least in my case the root cause seems to be quite different. In the same place of `RtApiAlsa :: getDeviceInfo` code quoted above, if `snd_ctl_open( &chandle, "default", SND_CTL_NONBLOCK )` is successfull and `device` is not zero, the `chandle` handle will never be closed. This results in that `unable to connect` error after a few open-close cycles. I also have to mention that this error happens even when you actually open the default (0) device. This is because Alsa api calls `getDeviceInfo` for all available devices in the `saveDeviceInfo` function.

The solution is simple - just add missing `snd_ctl_close`:
```
  result = snd_ctl_open( &chandle, "default", SND_CTL_NONBLOCK );
[...]

Issue #224, opened 2019-10-30
getDefaultOutputDevice returns wrong index on Ubuntu ALSA


witaly-iwanow:
I've modified playsaw sample a little to use ALSA instead of PulseAudio (PA works fine, but I need to switch audio devices which doesn't seem to be possible with PA+RtAudio), and it now it fails to detect the default output device (always takes 0, but it has to be the last one). Here's the log I get:
~/rtaudio/_cmake$ tests/playsaw 1 48000
num devices: 8
0 hw:HDA Intel HDMI,3
1 hw:HDA Intel HDMI,7
2 hw:HDA Intel HDMI,8
3 hw:HDA Intel HDMI,9
4 hw:HDA Intel HDMI,10
5 hw:HDA Intel PCH,0
6 hw:HDA Intel PCH,1
7 default
getDefaultOutputDevice returned 0

Playing ... press <enter> to quit (buffer size = 512).
RtAudio alsa: _NOT_ running realtime scheduling

<no sound>

"tests/playsaw 1 48000 7" works as expected
[playsaw.cpp.txt](https://github.com/thestk/rtaudio/files/3787030/playsaw.cpp.txt)

witaly-iwanow:
I'm on MacBook Pro 2015 something with Ubuntu 19.10 installed (real thing, not VM)

radarsat1:
I agree that there is a problem here, I was struggling with something similar yesterday.. I was getting errors opening the default device and found it was trying to open "hw:0" instead of "default", which didn't work because pulse audio was running.

I would like to look into it but don't have time at the moment unfortunately.  What should be the correct behaviour?  Should "default" be device 0?  When I do have time, I'll have to go back and re-read previous work on the alsa backend, because I remember this has been worked on before.

radarsat1:
I just committed dd23b6598b8e7f74d098628c852188a0d81e3562 which moves "default" to device 0.

In doing so however I've noticed that the way we enumerate devices is not really correct, at least with respect to how the "aplay" utility works that comes with ALSA.

In aplay, you can list "devices" (option `-l`) or "pcms" (option `-L`).  Our method corresponds more with "devices", where we iterate over devices and subdevices.  However the loop iterating over "pcms" is much simpler and includes more meaningful names.  Also, the `-D` option used to select the play device corresponds with the "PCM", or this latter iteration loop, not the one we use:

```
$ man aplay | grep -A1 -- --device
       -D, --device=NAME
              Select PCM by name
```

[...]

PR #296, opened 2021-05-25
Fix PulseAudio getDefaultOutputDevice() and getDefaultInputDevice()

### LINKED ISSUE #295: Incorrect default PulseAudio output device on some systems
On all Linux PulseAudio systems, calling `RtAudio::getDefaultOutputDevice()` with the PulseAudio backend returns device 0. On some systems, this is an input-only device which cannot be used as output, meaning the returned default output device cannot be used to open a stream. Why does this happen?

[RtApiPulse](https://github.com/thestk/rtaudio/blob/48d60dbc9f578df8552f6dfa000fc26bec1f8232/RtAudio.cpp#L8619-L8633) first calls `pa_context_get_sink_info_list()` to loop through all output devices and call `rt_pa_sink_info_cb`, which [inserts the default output device at index 0](https://github.com/thestk/rtaudio/blob/48d60dbc9f578df8552f6dfa000fc26bec1f8232/RtAudio.cpp#L8534-L8573). Then it calls `pa_context_get_source_info_list()` to loop through all input devices and call `rt_pa_source_info_cb`, which [inserts the default input device at index 0](https://github.com/thestk/rtaudio/blob/48d60dbc9f578df8552f6dfa000fc26bec1f8232/RtAudio.cpp#L8575-L8617)

`rt_pa_source_info_cb` runs second and attempts to insert the default input device at index 0. If it shares a name with an existing (eg. output) device, it [merges the two devices](https://github.com/thestk/rtaudio/blob/48d60dbc9f578df8552f6dfa000fc26bec1f8232/RtAudio.cpp#L8593-L8611). Otherwise, the default input device is inserted at index 0, moving the default output device to index 1.

But `RtApiPulse::getDefaultOutputDevice()` falls back to `RtApi::getDefaultOutputDevice()`, which is a placeholder which always returns 0. So on computers where the default input/output devices don't share the same name, `RtApiPulse::getDefaultOutputDevice()` doesn't return the correct default output device.

(Sidenote: I think that `RtApiPulse::getDefaultInputDevice()` also always returns index 0, which is incorrect if the default PulseAudio input device gets merged into a non-default PulseAudio output device, rather than being inserted at index 0.)

----

The idea I came up with was to loop over `RtApi::getDeviceInfo()` and return the (hopefully exactly 1) device labeled as a default. To my surprise, this code was already implemented... in RtApiWasapi only.

https://github.com/thestk/rtaudio/blob/48d60dbc9f578df8552f6dfa000fc26bec1f8232/RtAudio.cpp#L4526-L4548

In my local copy of RtAudio, I extracted this logic to `getDefaultInputDeviceImpl` and `getDefaultOutputDeviceImpl` functions and added overriding methods `RtApiPulse::getDefaultOutputDevice()` and `RtApiPulse::getDefaultInputDevice()` which delegated to the extracted looping functions. This fixes the behavior on PulseAudio on my machine.

I could upload my code, but it currently uses the C++11 `override` keyword and isn't stylistically consistent with the rest of RtAudio, so I don't want to send a PR yet. It seems RtAudio recently added a C++11 requirement (https://github.com/thestk/rtaudio/commit/78a48155a3d3af104e21d4843c36229dd780ddbf), ~~but only uses the `override` keyword inconsistently~~ I'd have to rebase to a newer RtAudio commit.

What I'm confused about is why the base class `RtApi::getDefaultOutputDevice()` doesn't loop over all devices by default. I have a hunch that making it do so would fix other audio backends as well. It appears that ALSA also has issues calculating the default device, looking at https://github.com/thestk/rtaudio/issues/224 and https://github.com/thestk/rtaudio/issues/270. However I haven't tested calling my new functions in ALSA, and I don't know if this new logic will fix/break/slow down other APIs aside from Pulse and ALSA. (WASAPI's behavior is unchanged, and DirectSound overrides `getDefaultOutputDevice` to always return device 0, even if the base class is changed. The other APIs will change their behavior.)

----

I noticed that RtApiPulse lacks support for `DUPLEX` PulseAudio. Is this a major issue for pro audio recording? (None of my programs record and play sound at the same time, so I can't judge the impact.) Do pro audio programs just not use PulseAudio in the first place?

nyanpasu64:
On all Linux PulseAudio systems, calling `RtAudio::getDefaultOutputDevice()` with the PulseAudio backend returns device 0. This is because `RtApiPulse::getDefaultOutputDevice()` is not defined and calling it invokes the base-class method `RtApi::getDefaultOutputDevice()`, which is a placeholder which always returns 0. But on computers where the default input/output devices don't share the same name, this is an input-only device which cannot be used as output, meaning the returned default output device cannot be used to open a stream.  (getDefaultInputDevice() may also be wrong in rare cases. Details on why these happen are at #295.)

The idea I came up with was to loop over `RtApi::getDeviceInfo()` and return the (hopefully exactly 1) device labeled as a default. To my surprise, this code was already implemented... in RtApiWasapi only.

https://github.com/thestk/rtaudio/blob/48d60dbc9f578df8552f6dfa000fc26bec1f8232/RtAudio.cpp#L4526-L4548

## Outdated

To fix this issue, I extracted this logic to `getDefaultInputDeviceImpl` and `getDefaultOutputDeviceImpl` functions and added overriding methods `RtApiPulse::getDefaultOutputDevice()` and `RtApiPulse::getDefaultInputDevice()` which delegated to the extracted looping functions. This fixes the behavior on PulseAudio on my machine.

What I'm confused about is why the base class `RtApi::getDefaultOutputDevice()` doesn't loop over all devices by default. I have a hunch that making it do so would fix other audio backends as well. It appears that ALSA also has issues calculating the default device, looking at https://github.com/thestk/rtaudio/issues/224 and https://github.com/thestk/rtaudio/issues/270. However I haven't tested calling my new functions in ALSA, and I don't know if this new logic will fix/break/slow down other APIs aside from Pulse and ALSA. (WASAPI's behavior is unchanged, and DirectSound overrides `getDefaultOutputDevice` to always return device 0, even if the base class is changed. The other APIs will change their behavior.)

## Update (2021-06-23)

I force-pushed the branch to change RtApi::getDefaultInputDevice() and RtApi::getDefaultOutputDevice() to properly query the default devices. This implementation is used by all RtApi subclasses unless they override the methods. Currently, RtApiDs (DirectSound) overrides both methods to return 0, and RtApiCore (Core Audio) overrides them with Mac-specific code.

Fixes #295.

### DIFF
--- a/RtAudio.cpp
+++ b/RtAudio.cpp
@@ -410,13 +410,25 @@ void RtApi :: openStream( RtAudio::StreamParameters *oParams,
 
 unsigned int RtApi :: getDefaultInputDevice( void )
 {
-  // Should be implemented in subclasses if possible.
+  // Should be reimplemented in subclasses if necessary.
+  for ( unsigned int i = 0; i < getDeviceCount(); i++ ) {
+    if ( getDeviceInfo( i ).isDefaultInput ) {
+      return i;
+    }
+  }
+
   return 0;
 }
 
 unsigned int RtApi :: getDefaultOutputDevice( void )
 {
-  // Should be implemented in subclasses if possible.
+  // Should be reimplemented in subclasses if necessary.
+  for ( unsigned int i = 0; i < getDeviceCount(); i++ ) {
+    if ( getDeviceInfo( i ).isDefaultOutput ) {
+      return i;
+    }
+  }
+
   return 0;
 }
 
@@ -4520,34 +4532,6 @@ Exit:
     error( errorType );
   return info;
 }
-
-//-----------------------------------------------------------------------------
-
-unsigned int RtApiWasapi::getDefaultOutputDevice( void )
-{
-  for ( unsigned int i = 0; i < getDeviceCount(); i++ ) {
-    if ( getDeviceInfo( i ).isDefaultOutput ) {
-      return i;
-    }
-  }
-
-  return 0;
-}
-
-//-----------------------------------------------------------------------------
-
-unsigned int RtApiWasapi::getDefaultInputDevice( void )
-{
-  for ( unsigned int i = 0; i < getDeviceCount(); i++ ) {
-    if ( getDeviceInfo( i ).isDefaultInput ) {
-      return i;
-    }
-  }
-
-  return 0;
-}
-
-//-----------------------------------------------------------------------------
 
 void RtApiWasapi::closeStream( void )
 {
--- a/RtAudio.h
+++ b/RtAudio.h
@@ -1027,8 +1027,6 @@ public:
   RtAudio::Api getCurrentApi( void ) override { return RtAudio::WINDOWS_WASAPI; }
   unsigned int getDeviceCount( void ) override;
   RtAudio::DeviceInfo getDeviceInfo( unsigned int device ) override;
-  unsigned int getDefaultOutputDevice( void ) override;
-  unsigned int getDefaultInputDevice( void ) override;
   void closeStream( void ) override;
   void startStream( void ) override;
   void stopStream( void ) override;


nyanpasu64:
Also CI is currently running on travis-ci.org, which is going down on 2021-06-15 (in a few days). It's probably necessary to migrate to travis-ci.com (limited free build minutes) or another service.

garyscavone:
I would prefer that the base class RtApi::getDefaultOutputDevice() and RtApi::getDefaultInputDevice() functions loop over all devices by default, rather than create new functions. I think it should work in general and I think it was just an oversight that it was not implemented that way in the past. The only issue I see with that behaviour is that there may be a delay between when getDeviceInfo() is called (and when the default devices are registered) and when the user calls getDefaultOutput/InputDevice(), during which the default device may have been modified in the system. But for APIs that don't allow the default devices to be queried dynamically, that is the best we can do.

nyanpasu64:
I force-pushed the branch to change RtApi::getDefaultInputDevice() and RtApi::getDefaultOutputDevice() to properly query the default devices. This implementation is used by all RtApi subclasses unless they override the methods. Currently, RtApiDs (DirectSound) overrides both methods to return 0, and RtApiCore (Core Audio) overrides them with Mac-specific code.

It works on PulseAudio, but I haven't tested other APIs.

nyanpasu64:
poke

nyanpasu64:
just realized `getDeviceCount()` is performing queries to pulseaudio (and probably other apis) on every single output device, and we call it once per loop iteration, resulting in redundant work, O(n^2) individual devices being probed, and possible tearing if devices are plugged or unplugged during the loop... I'll try changing this to cache the device count.

Is `getDeviceCount()` intended to re-query the device count rather than reusing the first cached value?

garyscavone:
Yes, because devices may come and go (hot-plugged), it is designed to recheck each time called.

On Jul 30, 2021, at 4:22 PM, nyanpasu64 ***@***.******@***.***>> wrote:



just realized getDeviceCount() is performing queries to pulseaudio (and probably other apis) on every single output device, and we call it once per loop iteration, resulting in redundant work, O(n^2) individual devices being probed, and possible tearing if devices are plugged or unplugged during the loop... I'll try changing this to cache the device count.

Is getDeviceCount() intended to re-query the device count rather than reusing the first cached value?

Issue #302, opened 2021-07-30
Pkgconfig file missing CoreAudio linking flags on macOS


ntonnaett:
This is the pkgconfig file you get when installing RtAudio with brew on macOS:
```
prefix=/usr/local/Cellar/rt-audio/5.1.0_1
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include/rtaudio        

Name: librtaudio
Description: RtAudio - a set of C++ classes that provide a common API for realtime audio input/output
Version: 5.1.0
Requires:  
Libs: -L${libdir} -lrtaudio
Libs.private: -lpthread
Cflags: -pthread -I${includedir}  -D__MACOSX_CORE__
```
The define says it's using the CoreAudio backend but it misses the linking flags for the CoreAudio framework. I think `-framework CoreAudio' has to be added to Libs.

PR #303, opened 2021-07-31
Let Meson generate pkgconfig file

ntonnaett:
This change lets Meson generate a pkgconfig file that looks like this (edit: removed defines):
```
prefix=/usr/local
libdir=${prefix}/lib64
includedir=${prefix}/include

Name: rtaudio
Description: RtAudio - a set of C++ classes that provide a common API for realtime audio input/output
Version: 5.1.0
Requires.private: alsa, jack, libpulse-simple
Libs: -L${libdir} -lrtaudio
Libs.private: -pthread
Cflags: -I${includedir}/rtaudio
```

It would be quite interesting if Meson adds apple frameworks to Libs.private. Then #302 would be fixed with that (at least for Meson). 

There are two things different to the rtaudio.pc.in file.
1. All dependencies are private because RtAudio doesn't expose any types or API from them.
2. The libdir for Fedora gets correctly set to lib64.

~~RTAUDIO_EXPORT seems to be a Windows thing but it's additionally checked in the header. Should I define that only in Windows?
HAVE_GETTIMEOFDAY is checked in the header. So I thought that would be correct.~~

Edit: After #308 is merged compiler defines shouldn't be necessary anymore. So I removed them completely from the pkgconfig file.

---

This PR also removes the direct libpulse dependency because libpulse-simple seems to be sufficient.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -78,7 +78,7 @@ if host_machine.system() == 'windows'
 	deps += compiler.find_library('winmm', required: true)
 endif
 
-rtaudio = library('rtaudio', src, include_directories: incdir, cpp_args: defines, dependencies: deps)
+rtaudio = library('rtaudio', src, include_directories: incdir, cpp_args: defines, dependencies: deps, install: true)
 rtaudio_dep = declare_dependency(include_directories : '.',
   link_with : rtaudio)
 meson.override_dependency('rtaudio', rtaudio_dep)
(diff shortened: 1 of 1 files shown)


ntonnaett:
I would like to merge #308 first and then remove all defines from Cflags.

ntonnaett:
I should probably add the soversion too.

PR #308, opened 2021-08-06
Check predefined macros instead of backend macros to determine the OS

ntonnaett:
This change uses predefined macros instead of audio backend macros to determine the OS. With this change we could remove the backend defines when using a precompiled RtAudio.


radarsat1:
I think this makes sense as it only affects the types used for threading, and has nothing to do with the actual audio APIs, so selecting based on the OS makes sense here.

radarsat1 (approved):

PR #310, opened 2021-08-07
Fix header problem with CMake package configuration file

### LINKED ISSUE #283: How to use with CMake / different include paths on unix/windows?
I have installed the lib with vcpkg and used it with
```cmake
find_path(RtAudio_INCLUDE_DIR rtaudio/RtAudio.h)
find_library(RtAudio_LIBRARY rtaudio)
target_include_directories(lichtsteuerung PRIVATE ${RtAudio_INCLUDE_DIR})
target_link_libraries(lichtsteuerung ${RtAudio_LIBRARY})
```
But the include dir is `rtaudio/RTAudio.h` on Windows and `RtAudio.h` on Linux/Mac. I have also tried
```cmake
find_package(RtAudio CONFIG REQUIRED)
target_link_libraries(lichtsteuerung RtAudio::RtAudio)
```
but that didn't worked.  
[Here](https://github.com/microsoft/vcpkg/blob/master/ports/rtaudio/portfile.cmake) you can find the `vcpkg` portfile, if you would say that this is a problem of vcpkg and not rtaudio.

albestro:
Fix #283

I just went for fixing this and I realized there was an issue already opened reporting the same problem.

I tested on macOS and it works. @autoantwort can you give it a try?

EDIT:
Just for completeness, the problem reported in the issue is not exactly the same but it is somehow correlated.
Indeed, I would suggest to use the project like this
```cmake
find_package(RtAudio)
add_exectuable(... PRIVATE RtAudio::rtaudio)
```
specifying the install root path in `RtAudio_ROOT` if needed (as env variable or cmake cache variable).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -251,8 +251,6 @@ if(MSVC AND RTAUDIO_STATIC_MSVCRT)
   endforeach(flag_var)
 endif()
 
-# Add headers destination for install rule.
-set_property(TARGET rtaudio PROPERTY PUBLIC_HEADER RtAudio.h rtaudio_c.h)
 set_target_properties(rtaudio PROPERTIES
   SOVERSION ${SO_VER}
   VERSION ${FULL_VER})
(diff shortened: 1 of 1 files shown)


albestro:
As soon as I have time I may give a look if it is easy to add a simple test for this kind of thing in the CI.

albestro:
If there is an agreement, I would also propose to remove this which may cause some headache.

https://github.com/thestk/rtaudio/blob/d77b5b642d8c66944ba158a3e0e1e2f91d5d3818/CMakeLists.txt#L305-L306

autoantwort:
I will try this next week

autoantwort:
LGTM 👍

autoantwort (approved):

PR #312, opened 2021-08-12
Fix pkgconfig file for linking statically with CoreAudio…

### LINKED ISSUE #302: Pkgconfig file missing CoreAudio linking flags on macOS

ntonnaett:
… and make all dependencies private.

As I don't have macOS I can only assume that this fixes #302


radarsat1:
Perfect, I just committed something similar to RtMidi regarding the `Require.private` thing.  would it be possible for you to check over on that project to see if a similar change for "framework" is necessary?

ntonnaett:
Yes, it's missing there too. I can make a PR.

PR #313, opened 2021-08-12
Fix check predefined macros

radarsat1:
A fix for some issues with #308.

### DIFF
--- a/RtAudio.cpp
+++ b/RtAudio.cpp
@@ -56,7 +56,7 @@ const unsigned int RtApi::SAMPLE_RATES[] = {
   32000, 44100, 48000, 88200, 96000, 176400, 192000
 };
 
-#if defined(__WINDOWS_DS__) || defined(__WINDOWS_ASIO__) || defined(__WINDOWS_WASAPI__)
+#if defined(_WIN32) || defined(__CYGWIN__)
   #define MUTEX_INITIALIZE(A) InitializeCriticalSection(A)
   #define MUTEX_DESTROY(A)    DeleteCriticalSection(A)
   #define MUTEX_LOCK(A)       EnterCriticalSection(A)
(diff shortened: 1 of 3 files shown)


RtAudio.cpp
```diff
@@ -82,15 +82,12 @@ const unsigned int RtApi::SAMPLE_RATES[] = {
     return s;
   }
 
-#elif defined(__LINUX_ALSA__) || defined(__LINUX_PULSE__) || defined(__UNIX_JACK__) || defined(__LINUX_OSS__) || defined(__MACOSX_CORE__)
+#elif defined(__unix__) || defined(__APPLE__)
```
falkTX, line 85: shouldn't this just be `#else` ?
There are non-unix POSIX systems that have pthreads support (HaikuOS is an example) and there is no fallback `#else` here anymore..

RtAudio.h
```diff
@@ -622,11 +622,14 @@ class RTAUDIO_DLL_PUBLIC RtAudio
   typedef pthread_t ThreadHandle;
   typedef pthread_mutex_t StreamMutex;
 
-#else // Setup for "dummy" behavior
+#endif
+
+// Setup for "dummy" behavior if no apis specified.
+#if !(defined(__WINDOWS_DS__) || defined(__WINDOWS_ASIO__) || defined(__WINDOWS_WASAPI__) \
+      || defined(__LINUX_ALSA__) || defined(__LINUX_PULSE__) || defined(__UNIX_JACK__) \
```
falkTX, line 629: shouldn't this be `RTAUDIO_API_PULSE`, `RTAUDIO_API_OSS` etc?

RtAudio.cpp
ntonnaett: Fair point. This should be changed for #308 too.

radarsat1: I generally agree with this, I don't think trying to enumerate any and all systems that support pthreads is necessarily the right approach and since there is no other fall-back, I guess simply `#else` could work.   Maybe in the future we can switch to C++11 threads and not worry about this kind of thing?  Care to make a PR @falkTX?

RtAudio.h
radarsat1: No, those are only CMake flags, not used in C++.

RtAudio.cpp
radarsat1: No worries, done.

falkTX: I think just changing it to always fallback to pthread is fine tbh, rtaudio would not function without it anyway.
Just need to be mentioned somewhere that this is a requirement.

Issue #212, opened 2019-08-29
RFC: new API w/o exceptions (partial implementation)


garyscavone:
I've created a new branch (noexceptions) with a modified RtAudio API to remove C++ exceptions and instead return error codes where appropriate. Currently, this is only implemented for the CoreAudio and JACK APIs and the audioprobe.cpp, playsaw.cpp and duplex.cpp test programs.

Before I spend the time to update the other APIs and test programs, I'd like feedback from the community as to whether this seems like a useful change and whether it should become the default once fully implemented. I also added the option to specify an errorCallback function in the RtAudio constructor.

radarsat1:
Can't believe it's been almost a year since you did this.. I was reviewing things and came across your branch, I'm sorry it completely slipped my mind.
Actually I find it improves the code quite a bit, as it removes try/catches everywhere that repeat the same code in the "catch".  (In practice it's rare to have a single exception handler that aborts the whole program.)

I've merged master into this branch and updated a few other APIs and the C backend to adapt to it.

By the way I was going to add a "userData" parameter to the error callback so that I could store error messages somewhere, and then realized that now with C++11 we can do something a little fancier with std::function.
https://github.com/thestk/rtaudio/blob/noexceptions/rtaudio_c.cpp#L66

nyanpasu64:
Not an RtAudio developer, but is it a good idea to mark `RtAudioErrorType` as `enum class` (C++11) and `[[nodiscard]]` (C++17)?

radarsat1:
I tried using an `enum class` and it's kind of okay, but the only annoying thing is that there is no way to convert to bool.  Currently with the value of `RTAUDIO_NO_ERROR == 0`, all the examples do `if ( dac.openStream() ) ...`

with `enum class`, every instance of this has to be changed to,

```
    if ( adc->openStream( &oParams, &iParams, RTAUDIO_SINT32, fs, &bufferFrames, &pulse, (void *)&mydata ) != RtAudioErrorType::NO_ERROR ) {
```

which is a bit more verbose.  I'm not completely against it but it's too bad it comes with this annoyance.  Apparently you can't write a bool cast operator for enum classes, which is too bad.

garyscavone:
Closing ... this has been completed across all APIs.

PR #322, opened 2021-09-15
No-exceptions branch

radarsat1:
This is @garyscavone's no-exceptions branch, updated to current master, that was well received in #212.  I propose to merge it now; it will affect current user code, so the other option is to wait until after the next release, but I prefer not to wait.  Open for comments, concerns about this idea.


radarsat1:
I should add, I have looked it over once more and don't find egregious issues.  However, no guarantee that it's problem free, but in my experience there's no better way to find out except putting it on master and seeing who complains.  It will never get completely vetted by all interested parties until it's released, so no point waiting. (imho)  It's also getting harder to maintain a diverging branch, so.. already it's difficult to "diff" this with master to figure out what the overall changes are.

radarsat1:
Note: this branch also contains a version bump.

ntonnaett:
Shouldn't those functions be made noexcept now?

garyscavone:
I know there are some bugs in this PR, particularly related to WASAPI, but I think it is easier to merge this first and then fix them after.

PR #327, opened 2021-10-11
ALSA: rewrite of device listing logic to be more similar to aplay

radarsat1:
Uses the same logic as the "aplay" utility that comes with ALSA to get the device list,
with some additional filtering for undesired plugins like "plughw" and "null".

Moreover, the list is saved on first use, or whenever `getDeviceCount()` is called, and
this list is thereafter used to index the device name, which should lead to stable device
selection behaviour.

When devices are initially listed, each device is open-probed to check
availability.  (ALSA errors due to this are suppressed, but it does result in some
messages from the Jack library if the ALSA jack plugin is installed.)

The result is a device list more similar to what is seen in popular software like
Audacity, etc.

I'm curious to hear from people willing to test this or give informed opinions about correct ALSA usage.

### DIFF
--- a/RtAudio.cpp
+++ b/RtAudio.cpp
@@ -48,6 +48,7 @@
 #include <climits>
 #include <cmath>
 #include <algorithm>
+#include <unordered_map>
 
 // Static variable definitions.
 const unsigned int RtApi::MAX_SAMPLE_RATES = 14;
@@ -7226,6 +7227,10 @@ static const char* getErrorString( int code )
 #include <alsa/asoundlib.h>
 #include <unistd.h>
 
+static std::unordered_map<std::string, void*> alsa_ignore_list = {
+  {"plughw", 0}, {"dsnoop", 0}, {"dmix", 0}, {"null", 0}, {"upmix", 0}, {"vdownmix", 0}
+};
+
   // A structure to hold various information related to the ALSA API
   // implementation.
 struct AlsaHandle {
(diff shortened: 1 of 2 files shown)


radarsat1:
I was just looking at this again.  I think I can try to come up with an alternative proposal, but actually I do think that, at least for ALSA, it's best to save the info like this.

It's not just about the textual names, it's actually more about the fact that ALSA keeps the list in a very sort of "noisy" linked list.  So there's not even a way to index that list without scanning it, and it tends to include a lot of names of "unopenable" devices that you wouldn't want to show in a user interface.   So the routine not only goes through a linked list of devices, but it probe-opens each device to get the device info (output/capture channels, etc) as well as the names, and in doing so it rejects a lot of undesired entries.

So, if you don't keep a list in memory, then when you want to open say, device 5, then you have to scan and probe-open maybe 10 or 20 devices before counting to 5.  It even causes non-suppressible error messages e.g. if Jack is not running (because ALSA has a Jack plugin).  So it's sort of much better to keep a list of those names and just index the list and open whichever one was selected, by giving the name that you saved from earlier.

That said, it is certainly feasible to change the code to freshly scan through the list during openStream, instead of keeping it around, I think it wouldn't even be a big code change.  So maybe I can propose that as an alternative.  I'm happy to keep the port selection problem separate.  But I do see my solution here as sort of helping to "normalize" ALSA towards the RtAudio numerical device model without (hopefully) breaking things and introducing the least amount of friction.

garyscavone:
A new device selection behaviour has been merged into master that addresses these issues, so I am closing this.

PR #332, opened 2021-11-09
ALSA: rewrite of device listing logic to be more similar to aplay (no-save alternative)

radarsat1:
Uses the same logic as the "aplay" utility that comes with ALSA to get the device list,
with some additional filtering for undesired plugins like "plughw" and "null".

Note that the list of devices is scanned and open-probed whenever `getDeviceCount()`,
`getDeviceInfo()`, or `openStream()` are called.

The result is a device list more similar to what is seen in popular software like
Audacity, etc.

This PR is presented as an alternative to #327.  Note that it results in much more "probing" out of necessity, since there is no information saved between calls to `getDeviceCount`, `geDeviceInfo`, and `openStream`; also, possible inconsistency if the computer's device list changes.

### DIFF
--- a/RtAudio.cpp
+++ b/RtAudio.cpp
@@ -48,6 +48,7 @@
 #include <climits>
 #include <cmath>
 #include <algorithm>
+#include <unordered_map>
 
 // Static variable definitions.
 const unsigned int RtApi::MAX_SAMPLE_RATES = 14;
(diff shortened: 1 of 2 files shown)


garyscavone:
I'm going to hold off merging this. If I run "aplay -l" on my current linux box, which has a built-in Intel soundcard and an RME hammerfall card, I get a clean list of 4 devices (1 hammerfall, Intel analog, Intel digital and Intel HDMI). If I run it with this PR, default and pulse are repeated and the card names and channels for both devices are vague and weird (10000 input and output channels). If I run it with the current master version of RtAudio, it is closer to what I see from aplay, albeit there is an unsuccessfully probed device at the end (not sure what that is). So, I think this needs work.

Ultimately, I might be in favour of moving toward a name-based device identification scheme, but I think we need to work on that more after a new release.

garyscavone:
Some of this PR was incorporated into the device selection updates, so I am closing this.

PR #352, opened 2022-05-08
handle a rare case where wasapi would not return the device name

divideconcept:
I encountered a rare case among my beta testers where Windows 10 would return 10 render devices when only 9 were publicly listed in Windows Sound Playback panel. That 10th device seemed to be a ghost device with no name and no user visibility.
As a result, when RtAudio was querying the 10th device name, devicePropStore->GetValue( PKEY_Device_FriendlyName, &deviceNameProp ) returned no error, but deviceNameProp.pwszVal was set to nullptr, which then leaded to a crash.

Making sure that deviceNameProp.pwszVal is not null avoid that crash.
I also added a similar check for defaultDeviceNameProp.pwszVal.

For reference all the details about that crash are in that private beta forum: https://forums.steinberg.net/t/spectralayers-9-0-0-283-not-launching-windows-10/784659/24


MarcusTomlinson:
Awesome, thanks for the fix!

PR #375, opened 2022-11-12
meson: only build pulseaudio under linux

neheb:
Fixes compilation under MSYS2. Needs work to be compatible.

Missing pthread.h, mutex wrong type, and maybe other stuff.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -69,7 +69,8 @@ if alsa_dep.found()
 endif
 
 jack_dep = dependency('jack', required: get_option('jack'))
-if jack_dep.found()
+# broken on MSYS2. Need to use pthreads instead of Windows threading
+if jack_dep.found() and host_machine.system() != 'windows'
 	defines += '-D__UNIX_JACK__'
 	deps += jack_dep
 endif
@@ -79,7 +80,8 @@ if get_option('oss') == true
 endif
 
 pulsesimple_dep = dependency('libpulse-simple', required: get_option('pulse'))
-if pulsesimple_dep.found()
+# broken on MSYS2. Need to use pthreads instead of Windows threading
+if pulsesimple_dep.found() and host_machine.system() != 'windows'
 	defines += '-D__LINUX_PULSE__'
 	deps += pulsesimple_dep
 endif


ntonnaett:
This will probably break BSD. You can turn pulseaudio off with `meson configure -Dpulse=disabled` in your build directory or use the same flag at the setup stage. Why is MSYS2 offering pulseaudio, if it's unusable on Windows?

neheb:
It's usable, just the code here needs adjustment.

neheb:
Done. CI errors are unrelated.

ntonnaett (approved):

ntonnaett:
Thanks for improving this on Windows! I've wrote the windows part of the meson build just with CI errors. I'm really happy it is useful.

neheb:
Just found out same problem exists with jack.

ntonnaett:
Then we should fix the actual bugs. JACK is actual useful on Windows. Although it would be good to add some weakjack support.

neheb:
I tried and was unsuccessful. JACK and PulseAudio need pthreads while DirectSound/WASAPI need Windows threading.

neheb:
```
--- a/RtAudio.cpp
+++ b/RtAudio.cpp
@@ -56,7 +56,7 @@ const unsigned int RtApi::SAMPLE_RATES[] = {
   32000, 44100, 48000, 88200, 96000, 176400, 192000
 };

-#if defined(_WIN32) || defined(__CYGWIN__)
+#if defined(_MSC_VER)
   #define MUTEX_INITIALIZE(A) InitializeCriticalSection(A)
   #define MUTEX_DESTROY(A)    DeleteCriticalSection(A)
   #define MUTEX_LOCK(A)       EnterCriticalSection(A)
@@ -82,12 +82,30 @@ const unsigned int RtApi::SAMPLE_RATES[] = {
     return s;
   }

-#elif defined(__unix__) || defined(__APPLE__)
+#else
   // pthread API
   #define MUTEX_INITIALIZE(A) pthread_mutex_init(A, NULL)
   #define MUTEX_DESTROY(A)    pthread_mutex_destroy(A)
   #define MUTEX_LOCK(A)       pthread_mutex_lock(A)
   #define MUTEX_UNLOCK(A)     pthread_mutex_unlock(A)
+
+  template<typename T> inline
+  std::string convertCharPointerToStdString(const T *text);
+
+  template<> inline
+  std::string convertCharPointerToStdString(const char *text)
+  {
+    return std::string(text);
+  }
+
+  template<> inline
+  std::string convertCharPointerToStdString(const wchar_t *text)
+  {
+    std::wstring ws(text);
+    std::string s(ws.begin(), ws.end());
+    return s;
+  }
+
 #endif

 // *************************************************** //
diff --git a/RtAudio.h b/RtAudio.h
index a7853f9..ba644a4 100644
--- a/RtAudio.h
+++ b/RtAudio.h
@@ -630,7 +630,7 @@ class RTAUDIO_DLL_PUBLIC RtAudio
 };

 // Operating system dependent thread functionality.
-#if defined(_WIN32) || defined(__CYGWIN__)
+#if defined(_MSC_VER)

   #ifndef NOMINMAX
     #define NOMINMAX

```

Is enough to compile. Probably doesn't work correctly though. AFAIK, the wchar_t needs to be converted from UTF-16 to UTF-8.

neheb:
BTW why isn't this project using std::thread and other C++11 stuff?

ntonnaett:
I see you found the same. MSYS2, MinGW and CYGWIN seem that they should all use pthreads. But if RtAudio is used with msvc or clang in a windows environment, it shouldn't. Your patch goes in the right direction but I think it's not quite right.

Some information about MSYS2 predefined macros.
https://www.msys2.org/wiki/Porting/

MSys2 has different environments. This was new to me. 
https://www.msys2.org/docs/environments/

The wchar conversion is probably a windows thing. I think this change needs more eyes and tests on different platforms. It's late here. Can't wrap my head around right now.

RtAudio was originally written years before C++11. std::thread could clean things up. Maybe ask the maintainers before if you want to work on this.

neheb:
clang-cl and msvc both define MSC_VER so it should be fine.

As far as std::thread conversion, I don't really see much commit activity. Would make sense to do other cleanups first.

The patch above is mostly correct with the convert function needing to use codecvt.

ntonnaett:
Nice. Do you want to open another PR for this?

PR #378, opened 2022-12-09
use pthreads on MinGW and Cygwin

neheb:
Needed for PulseAudio and JACK.

Signed-off-by: Rosen Penev <rosenp@gmail.com>

ping @ntonnaett

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -178,9 +178,9 @@ AS_IF([test "x$systems" = "x"],
     [*-*-freebsd*],  [systems="oss"],
     [*-*-linux*],    [systems="alsa pulse jack oss"],
     [*-apple*],      [systems="core jack"],
-    [*-mingw32*],    [systems="asio dsound wasapi jack"],
-    [*-mingw64*],    [systems="asio dsound wasapi jack"],
-    [*-msys*],       [systems="asio dsound wasapi jack"],
+    [*-mingw32*],    [systems="asio dsound pulse wasapi jack"],
+    [*-mingw64*],    [systems="asio dsound pulse wasapi jack"],
+    [*-msys*],       [systems="asio dsound pulse wasapi jack"],
   ))
 
 # If any were specifically requested disabled, do it.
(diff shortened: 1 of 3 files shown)


configure.ac
```diff
@@ -178,9 +178,9 @@ AS_IF([test "x$systems" = "x"],
     [*-*-freebsd*],  [systems="oss"],
     [*-*-linux*],    [systems="alsa pulse jack oss"],
     [*-apple*],      [systems="core jack"],
-    [*-mingw32*],    [systems="asio dsound wasapi jack"],
-    [*-mingw64*],    [systems="asio dsound wasapi jack"],
-    [*-msys*],       [systems="asio dsound wasapi jack"],
+    [*-mingw32*],    [systems="asio dsound pulse wasapi jack"],
```
ntonnaett, line 181: Is there a translation layer for pulseaudio in mingw?

ntonnaett: Sorry, I just forgot what this was all about.

configure.ac
```diff
@@ -193,6 +193,7 @@ AS_IF([test "x$with_asio"   = "xno"], [systems=`echo $systems|tr ' ' \\\\n|grep
 AS_IF([test "x$with_dsound" = "xno"], [systems=`echo $systems|tr ' ' \\\\n|grep -v dsound`])
 AS_IF([test "x$with_wasapi" = "xno"], [systems=`echo $systems|tr ' ' \\\\n|grep -v wasapi`])
 systems=" `echo $systems|tr \\\\n ' '` "
+need_pthread=yes
```
ntonnaett, line 196: I'm not a Windows user. But I think especially use cases without mingw and msys2 should be tested for regressions. Were WASAPI et al compiled without pthread before? Are they still working? Should this need_pthread be put in an if clause for Pulseaudio and JACK?

However RtAudio development is slow and there should be plenty of time to fix problems that occur with this change.

Just saw that you also cleaned up the wchar conversion. Nice.

ntonnaett:
@garyscavone What do you think?

garyscavone:
I can't check these changes easily myself but it looks like all the github checks have passed. The only thing I don't understand is that it is saying that this branch has conflicts that must be resolved (in RtAudio.cpp).

ntonnaett:
What I've checked out with the github cli is not just one commit beyond master. Interesting. @neheb Could you try to rebase?

neheb:
Rebased

PR #382, opened 2022-12-20
Fix audio glitch on Linux + PulseAudio.

cosinekitty:
This change resolves clicking and popping that occurs when using rtaudio to render output audio via PulseAudio on Linux systems. I discovered this issue using [VCV Rack](https://github.com/VCVRack/Rack) versions 2.1.0 and above, which uses a fork of rtaudio. [Andrew Belt](https://github.com/AndrewBelt) suggested I send this pull request upstream to the original rtaudio repo, because it will likely help more people that way.

Code to demonstrate reproducing the glitch and validating the fix, along with details, are available at:

https://github.com/cosinekitty/pulseaudio_glitch_demo

I will be happy to provide more help and info as needed.

Thank you!

### DIFF
--- a/RtAudio.cpp
+++ b/RtAudio.cpp
@@ -9456,22 +9456,8 @@ bool RtApiPulse::probeDeviceOpen( unsigned int deviceId, StreamMode mode,
     }
     break;
   case OUTPUT: {
-    pa_buffer_attr * attr_ptr;
-
-    if ( options && options->numberOfBuffers > 0 ) {
-      // pa_buffer_attr::fragsize is recording-only.
-      // Hopefully PortAudio won't access uninitialized fields.
-      buffer_attr.maxlength = bufferBytes * options->numberOfBuffers;
-      buffer_attr.minreq = -1;
-      buffer_attr.prebuf = -1;
-      buffer_attr.tlength = -1;
-      attr_ptr = &buffer_attr;
-    } else {
-      attr_ptr = nullptr;
-    }
-
     pah->s_play = pa_simple_new( NULL, streamName.c_str(), PA_STREAM_PLAYBACK,
-                                 dev_output, "Playback", &ss, NULL, attr_ptr, &error );
+                                 dev_output, "Playback", &ss, NULL, NULL, &error );
     if ( !pah->s_play ) {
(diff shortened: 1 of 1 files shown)


dromer:
@cosinekitty  The project in your url does not exist, or is not public.

cosinekitty:
> @cosinekitty The project in your url does not exist, or is not public.

Oops! I just changed it to be public. Thank you for letting me know.

garyscavone:
It seems that the attr_ptr structure allows one to refine aspects of buffer that might be useful to some people. And it is only used if someone specifies the numberOfBuffers parameter as an option. Did you try changing the values of numberOfBuffers to assess what was going on? If you use a very low number, it is not surprising that there may be some glitching when switching windows.

cosinekitty:
OK, I did some more experiments tonight. I added some extra code to my demo so when I press ENTER, it intentionally causes a click to the audio. That way I can get a sense of latency. I can press ENTER and hear if there is a perceptible time delay before the click.

I found that my suggested change to RtAudio.cpp, i.e. passing NULL instead of attr_ptr to pa_simple_new, noticeably increases latency. It sounds like an extra quarter second or half second, something like that. I don't think VCV Rack musicians are going to like that, especially when using MIDI keyboards, etc.

So I reverted my changes to RtAudio.cpp. Now it matches your master branch. As expected, I confirmed the glitches still happen with numberOfBuffers = 2. There are fewer of them when numberOfBuffers = 3. When I set it to 4, I haven't yet heard any glitches, plus the latency is imperceptible.

@AndrewBelt  So the evidence now is that it would be better to fix this in VCV Rack by either
* Changing [this line](https://github.com/VCVRack/Rack/blob/43cc6d02ff761f78a43d2a54c0dc8a46a44a50de/src/rtaudio.cpp#L97) to set numberOfBuffers = 4
* or perhaps adding a configuration setting where people experiencing the problem (like I do) can set numberOfBuffers to whatever we want, but leave the default to 2, to avoid any risk to existing users on platforms other than Linux + PulseAudio.

garyscavone:
This does not seem like an RtAudio issue (or there is a way to fix it by specifying the number of buffers to use) ... closing.

PR #384, opened 2022-12-26
Fixed leaking of WASAPI IAudioClient3 interface

mvshyvk:
Fixed leaking of IAudioClient3 interface. Leds on USB audio devices kept flashing even after closing the stream and freeing RtApi object.


mvshyvk:
@garyscavone Please, review fix for leaked interface in WASAPI implementation

PR #393, opened 2023-04-01
Asio probe infinite-loop fix.

dbadb:
The problem: when my ASIO device is unplugged (driver still installed), deviceProbe finds no devices which results in an infinite loop when `RtApiAsio :: probeDevices` invokes getDefaultInputDevice() - which re-invokes  RtApiAsio :: probeDevices.  This problem can be reproduced by running the standard audioprobe test.

In this condition `probeDeviceInfo( info ) == false` around line 3406.

I'm not claiming that my fix is the best fix but have verified that it unclogs the logjam.

### DIFF
--- a/RtAudio.cpp
+++ b/RtAudio.cpp
@@ -3370,7 +3370,6 @@ RtApiAsio :: ~RtApiAsio()
 void RtApiAsio :: probeDevices( void )
 {
   // See list of required functionality in RtApi::probeDevices().
-
   unsigned int nDevices = drivers.asioGetNumDev();
   if ( nDevices == 0 ) {
     deviceList_.clear();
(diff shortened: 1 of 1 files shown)


garyscavone:
I see extraneous code and comments (not currently in the master branch) from an earlier attempt to fix the issue. I understand the fix to the problem is to only call the "getDefaultXXDevice()" functions if the size of deviceList_ > 0?

dbadb:
indeed, this is the heart of it:

```c++
if(deviceList_.size() > 0)
  {
    getDefaultInputDevice();
    getDefaultOutputDevice();
  }
```

dbadb:
This was an attempt to squash the multi-step revision to reduce noise of extraneous changes. 
(git rebase newbie here)

marcelblum:
thank you so much for this, was driving me nuts 🥲

PR #426, opened 2024-04-06
Fix Windows deprecation warnings using WideCharToMultiByte()

g40:
This PR is for review. It is not final. Please add any comments and I'll adjust accordingly.


g40:
My bad. No Linux/MacOS CI on my Windows development systems.

PR #435, opened 2024-09-05
Fix no output on WASAPI (#433)

### LINKED ISSUE #433: WASAPI broken on master, callback called with 20k Hz rate
I'm getting no sound through WASAPI. The callback function is called too fast (20k times / sec), without waiting for the buffer to be played. There is no sound output, but it might be caused by my implementation not being able to deliver 20k buffers / sec.

I tried to bisect it, and I think it's caused by this PR (last commit on master): https://github.com/thestk/rtaudio/pull/431

The commit before works without any issues: https://github.com/thestk/rtaudio/commit/c5993f01df32421d9fd6bdf693cf0fa110613f89

lalitshankarchowdhury:
Fixes #433.

The issue was caused by creating new copies of ComPtr<IAudioClient> inside WasapiHandle instead of using references to the existing instance.

This likely led to improper management of the COM object's reference count, resulting in no audio output.

### DIFF
--- a/RtAudio.cpp
+++ b/RtAudio.cpp
@@ -52,8 +52,6 @@
 #include <locale>
 
 #if defined(_WIN32)
-#include <wrl/client.h>
-using Microsoft::WRL::ComPtr;
 #include <windows.h>
 #endif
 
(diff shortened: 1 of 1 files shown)


lalitshankarchowdhury:
Hi @garyscavone, could you review this when you have the chance? Thanks!

PR #445, opened 2024-10-31
Replaced deprecated string conversion function

ERSUCC:
When building RtAudio on Mac with C++17, a deprecation warning is thrown for the `std::wstring_convert` class. This pull request replaces that usage with conversion functions that have not been deprecated.

### DIFF
--- a/RtAudio.cpp
+++ b/RtAudio.cpp
@@ -45,10 +45,10 @@
 #include <iostream>
 #include <cstdlib>
 #include <cstring>
+#include <cwchar>
 #include <climits>
 #include <cmath>
 #include <algorithm>
-#include <codecvt>
 #include <locale>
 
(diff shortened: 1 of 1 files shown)


sonoro1234:
set(CMAKE_CXX_STANDARD 11) in CMakeLists.txt is not c++17.

ERSUCC:
@sonoro1234 I'm aware that the main repo currently uses C++11, but I have an ongoing personal project using C++17, and I am compiling RtAudio statically for that project. It's not a huge deal since as you said, C++11 is not deprecated, I just figured it is better to future-proof as long as the solution still works in C++11. Feel free to drop the pull request if you disagree, it still works as intended in my project.

sonoro1234:
You did not said it works with c++11!!
That is good.

PR #451, opened 2025-01-20
Switched audio callback from function pointer to std::function

ERSUCC:
This change allows member functions to be used as audio callbacks with std::bind, which can reduce the amount of user data that needs to be passed into the callback as an input.

### DIFF
--- a/RtAudio.cpp
+++ b/RtAudio.cpp
@@ -760,7 +760,7 @@ RtAudioErrorType RtApi :: openStream( RtAudio::StreamParameters *oParams,
       return error( RTAUDIO_SYSTEM_ERROR );
   }
 
-  stream_.callbackInfo.callback = (void *) callback;
+  stream_.callbackInfo.callback = callback;
   stream_.callbackInfo.userData = userData;
 
   if ( options ) options->numberOfBuffers = stream_.nBuffers;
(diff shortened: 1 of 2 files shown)


sonoro1234:
Given that the arguments are the same, I cant see how it would "reduce the amount of user data that needs to be passed into the callback as an input.". Perhaps an example would help. 
Also: C binding would be more complicated.

ERSUCC:
Hi Victor,

In my project, I had a bunch of audio-related state saved in a class. With the original C-style function pointers, I was not able to put the audio callback in the class as well. To get the data from the class to the callback, I had to put it into a struct and pass it into the user data parameter of the audio callback, and then I had to recreate the struct from void* in the callback itself. With std::function, I was able to bind a member function as the callback, and I was then able to directly access the audio state data without needing to use the user data parameter. It's not a huge deal, but the code is slicker and more expandable since I don't have to duplicate the necessary state in a separate user data struct. Hopefully that helped, but let me know if it still doesn't make sense; I may be misunderstanding something.

In regards to C binding: I did a quick test with the new std::function callback code and I was still able to use my old C-style callback code without changing anything. Did I just encounter a special case? I only made this PR because it seemed like it wouldn't affect any code using C-style function pointers. Also, since RtAudioErrorCallback is already an std::function, so it would make sense for both callbacks to match.

PR #452, opened 2025-01-21
proxy callback delete

sonoro1234:
I dont see this proxy callback necessary. In fact it works perfectly without it. I am missing something?


garyscavone:
I don't use the C API. The proxy callback generally looks unnecessary but I'll wait to see if someone else has some insight.

PR #454, opened 2025-02-20
New Feature: Buffer size and sample rate callback functions

faressc:
Some native APIs allow for dynamic buffer size and sample rate changes (i.e. JACK/Pipewire-JACK). This PR adds the ability to set callback functions for buffer size and sample rate changes that occur during stream operation. The callback functions are specified in the `RtAudio::StreamOptions` struct and are called when the buffer size or sample rate changes. The callback functions are backend agnostic and are implemented for the JACK backend in this PR.

Internally the `RtApiStream` gets two new members, `bufferSizeCallbackInfo` and `sampleRateCallbackInfo` similar to the existing `callbackInfo` member. The `BufferSizeCallbackInfo` and `SampleRateCallbackInfo` structs are populated when the `RtAudio::openStream` function is called. The user can specify the respective callback functions pointers (`RtAudioBufferSizeCallback` or/and an `RtAudioSampleRateCallback`) and user data pointers (`void*`) in the `RtAudio::StreamOptions` struct. The `RtAudio::StreamOptions` are therefore extended with the following members:

- `bufferSizeCallback` of type `RtAudioBufferSizeCallback`
- `bufferSizeCallbackUserData` of type `void*`
- `sampleRateCallback` of type `RtAudioSampleRateCallback`
- `sampleRateCallbackUserData` of type `void*`

When specified, the respective callback functions are called, when the buffer size or sample rate changes. Additionally, the following internal `RtApiStream` members are updated when the buffer size changes:

- `bufferSize`
- `userBuffer` is reallocated with the new buffer size
- `deviceBuffer` is reallocated with the new buffer size

When the sample rate changes, only the `sampleRate` member is updated.

This is all implemented with the JACK backend, but the changes are designed to be backend agnostic, so that other backends can be extended in the future.

I would appreciate feedback and would be happy to make any necessary changes to get this PR merged.

Cheers,
Fares

PS: I have also added the ability for the JACK backend to force a buffer size on the JACK server with `jack_set_buffer_size()` when the stream is opened and the requested buffer size from the stream is different from the current JACK server buffer size. This happens automatically, but a warning is printed to the console. To me this seems better than the current behavior where the buffer size of the JACK server is queried and the stream is simply opened with the JACK server buffer size.

### DIFF
--- a/RtAudio.cpp
+++ b/RtAudio.cpp
@@ -3370,83 +3370,134 @@ bool RtApiJack :: bufferSizeCallbackEvent ( unsigned long nframes )
 {
   if ( stream_.bufferSize == (unsigned int) nframes ) return true;
 
-  bool isRunning = stream_.state == STREAM_RUNNING;
-  if ( isRunning ) {
-    jack_client_t *client = ((JackHandle *) stream_.apiHandle)->client;
-    std::cout << "bufferSizeCallbackEvent: stopping stream." << std::endl;
-    jack_deactivate( client );
-    std::cout << "Client deactivated" << std::endl;
-    stream_.state = STREAM_STOPPED;
-  }
-  // Set the new buffer size
+  JackHandle *handle = (JackHandle *) stream_.apiHandle;
+  BufferSizeCallbackInfo *bufferSizeCallbackInfo = (BufferSizeCallbackInfo *) &stream_.bufferSizeCallbackInfo;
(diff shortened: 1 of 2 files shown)


garyscavone:
Would you consider writing a small test program for the tests directory to test this functionality? The added functionality seems great but could be challenging to test if added incrementally in other APIs.

PR #457, opened 2025-04-05
Return string by const ref to save a copy

autoantwort:
One could return the string by const ref or by value, but by const value does not make sense.

### DIFF
--- a/RtAudio.h
+++ b/RtAudio.h
@@ -589,7 +589,7 @@ class RTAUDIO_DLL_PUBLIC RtAudio
     non-zero RtAudioErrorType is returned by a function. This is the
     same message sent to the user-provided errorCallback function.
   */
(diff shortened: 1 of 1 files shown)


RtAudio.h
```diff
@@ -589,7 +589,7 @@ class RTAUDIO_DLL_PUBLIC RtAudio
     non-zero RtAudioErrorType is returned by a function. This is the
     same message sent to the user-provided errorCallback function.
   */
-  const std::string getErrorText( void );
+  const std::string &getErrorText(void);
```
autoantwort, line 592: I realized that clang-format changed the formatting a little bit. Do you have a `.clang-format` file that you use to format the code?

garyscavone: No.

PR #463, opened 2025-10-24
Add support for C++20 modules

mikomikotaishi:
I made a similar pull request [here](https://github.com/thestk/rtmidi/pull/357) on RtMidi, and am just applying the same edits here.

This pull request adds support for C++20 modules, as well as puts all symbols in the inlined namespace `rt::audio` (it is broken into nested namespace to emphasise the relation with RtMidi, which also shares namespace `rt`). This change does not break any existing code, but allows symbols to be optionally accessed through the namespace `rt::audio` in addition to the global namespace. The module itself is imported using `import rt.audio;` and is enabled with the build macro `RTAUDIO_BUILD_MODULES`. Enabling `RTAUDIO_USE_NAMESPACE` will make namespaces mandatory in the module.

Introducing modules to the project will not break any code, but is a benefit to users of C++20 and later who may wish to consume the library using modules rather than headers.

### DIFF
--- a/RtAudio.cpp
+++ b/RtAudio.cpp
@@ -55,9 +55,6 @@
 #include <windows.h>
 #endif
 
-inline namespace rt {
-inline namespace audio {
-
 // Static variable definitions.
 const unsigned int RtApi::MAX_SAMPLE_RATES = 14;
 const unsigned int RtApi::SAMPLE_RATES[] = {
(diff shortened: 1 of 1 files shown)


mikomikotaishi:
I'm not really sure what's causing the CI error for MinGW. If someone is more experienced and can explain it, that would be appreciated.

Issue #469, opened 2025-11-24
Nested namespace definitions need C++17


giraf-fe:
The latest pr changes the old:
```c++
inline namespace rt {
inline namespace audio { // ...
```
to:
```c++
namespace rt::audio { // ...
```
in RtAudio.h

However, the CMakeLists specifies the C++ standard as 11, while nested namespace definitions requires C++17. This is causing my build to fail.

I guess the two options would be to use the old way to specifiy the namespaces, or change the C++ version to 17. Simply changing the C++ version to 17 lets the library compile sucessfully. Not sure if projects still on 11 or 14 would be affected.

giraf-fe:
wait, made a mistake

giraf-fe:
Reopening, can reproduce using CMake and MSVC on WIndows with a fresh clone. Changing C++ version to 17 doesn't seem to fix either.

PR #470, opened 2025-11-25
remove nested namespace definitions to maintain compatibility with C++11

giraf-fe:
in reference to issue #469 

mentioning @mikomikotaishi since it reverts part of your change


mikomikotaishi:
Thanks for catching this. I believe RtMidi has the same problem as well.

PR #479, opened 2026-03-11
cmake: skip install code when embedded in other cmake build

timblechmann:
This is helpful when trying to embed rtaudio in e.g. via CPM/FetchContent.


CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 # Set minimum CMake required version for this project.
-cmake_minimum_required(VERSION 3.13 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.21 FATAL_ERROR)
```
timblechmann, line 2: compare https://cmake.org/cmake/help/latest/variable/PROJECT_IS_TOP_LEVEL.html ... it's not entirely necessary, though

PR #480, opened 2026-04-03
Resolve variable length array warnings when compiling with Clang

ERSUCC:
When compiling RtAudio on Mac with Clang, it threw a few warnings about arrays being allocated with variable length. Converting these stack arrays to heap arrays with `malloc` resolved the warnings.

### DIFF
--- a/RtAudio.cpp
+++ b/RtAudio.cpp
@@ -91,7 +91,7 @@ std::string convertCharPointerToStdString(const wchar_t* text)
   return nret;
 #else
   std::string result;
-  char dest[MB_CUR_MAX];
+  char* dest = (char*)malloc(sizeof(char) * MB_CUR_MAX);
   // get number of wide characters in text
   const size_t length = wcslen(text);
   for (size_t i = 0; i < length; i++) {
@@ -103,6 +103,7 @@ std::string convertCharPointerToStdString(const wchar_t* text)
       result += dest[j];
     }
   }
+  free(dest);
   return result;
 #endif
 }
@@ -1140,10 +1141,18 @@ void RtApiCore :: probeDevices( void )
     return;
   }
 
-  AudioDeviceID ids[ nDevices ];
+  AudioDeviceID* ids = (AudioDeviceID*)malloc(sizeof(AudioDeviceID) * nDevices);
+
+  if ( ids == NULL ) {
+    errorText_ = "RtApiCore::probeDevices: memory error allocating AudioDeviceID list.";
+    error( RTAUDIO_WARNING );
+    return;
+  }
+
   property.mSelector = kAudioHardwarePropertyDevices;
-  result = AudioObjectGetPropertyData( kAudioObjectSystemObject, &property, 0, NULL, &dataSize, (void *) &ids );
+  result = AudioObjectGetPropertyData( kAudioObjectSystemObject, &property, 0, NULL, &dataSize, (void *) ids );
   if ( result != noErr ) {
+    free(ids);
     errorText_ = "RtApiCore::probeDevices: OS-X system error getting device IDs.";
     error( RTAUDIO_SYSTEM_ERROR );
     return;
@@ -1181,6 +1190,8 @@ void RtApiCore :: probeDevices( void )
       deviceList_.erase( deviceList_.begin() + distance(deviceIds_.begin(), it ) );
     }
   }
+
+  free(ids);
 
   // Get default devices and set flags in deviceList_.
   AudioDeviceID defaultOutputId, defaultInputId;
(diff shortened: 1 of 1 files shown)


RtAudio.cpp
```diff
@@ -91,7 +91,7 @@ std::string convertCharPointerToStdString(const wchar_t* text)
   return nret;
 #else
   std::string result;
-  char dest[MB_CUR_MAX];
+  char* dest = (char*)malloc(sizeof(char) * MB_CUR_MAX);
```
nmq443, line 94: Should we use std::string instead of char* here? So that RAII will handle freeing resource for us, instead of using malloc()/free() manually

RtAudio.cpp
```diff
@@ -1140,10 +1141,18 @@ void RtApiCore :: probeDevices( void )
...
 
-  AudioDeviceID ids[ nDevices ];
+  AudioDeviceID* ids = (AudioDeviceID*)malloc(sizeof(AudioDeviceID) * nDevices);
+
+  if ( ids == NULL ) {
+    errorText_ = "RtApiCore::probeDevices: memory error allocating AudioDeviceID list.";
+    error( RTAUDIO_WARNING );
+    return;
+  }
+
```
nmq443, line 1151: Should we use std::vector instead of manually using malloc() ? Same reason as above

RtAudio.cpp
```diff
@@ -1140,10 +1141,18 @@ void RtApiCore :: probeDevices( void )
...
   property.mSelector = kAudioHardwarePropertyDevices;
-  result = AudioObjectGetPropertyData( kAudioObjectSystemObject, &property, 0, NULL, &dataSize, (void *) &ids );
+  result = AudioObjectGetPropertyData( kAudioObjectSystemObject, &property, 0, NULL, &dataSize, (void *) ids );
```
nmq443, line 1153: In my opinion, the reason we should use RAII instead because AudioObjectGetPropertyData() may throw (I don't know if it is marked as noexcept or not, please correct me if I am wrong), and if it does throw, the code won't run free() after this line of code, which lead to memory leak

RtAudio.cpp
```diff
@@ -1351,9 +1362,17 @@ bool RtApiCore :: probeDeviceInfo( AudioDeviceID id, RtAudio::DeviceInfo& info )
   }
 
   UInt32 nRanges = dataSize / sizeof( AudioValueRange );
-  AudioValueRange rangeList[ nRanges ];
-  result = AudioObjectGetPropertyData( id, &property, 0, NULL, &dataSize, &rangeList );
+  AudioValueRange* rangeList = (AudioValueRange*)malloc(sizeof(AudioValueRange) * nRanges);
```
nmq443, line 1365: Same as above

nmq443:
Hello @ERSUCC 
I know it is a little late but I have added some comments about your changes. I don't have a Mac so I don't know if my comments will help or not. Please correct me if I am wrong.
Thanks

ERSUCC:
@nmq443 Thank you for your comments! None of those functions are marked as `noexcept`, so you're right that there could be a memory leak. However, I don't think using `std::vector` is a good solution for any of these cases, because all of them are out parameters. It is possible to get the underlying pointer to a `std::vector` with `.data()`, but it feels kind of wrong to let an external function write to that underlying buffer instead of using the methods in the `std::vector` class, and I don't think that's the intended use of `.data()`. Also, I'm not a fan of using `std::vector` when the size of the array is known and constant; it just adds extra infrastructure.

My preferred solution would be to catch exceptions from these external functions, `free` the `malloc`'d array, and rethrow the exception. But also, if an exception occurs that prevents RtAudio from being initialized, there isn't much hope for the user's program anyway, so I doubt that this memory leak would be consequential in the first place. Either way, I'm happy to amend my PR with `try/catch` as necessary, if that seems like a good solution.

nmq443:
@ERSUCC Your comment about `std::vector::data()` sounds reasonable to me. Also I have just doing some research, using `.data()` to let a C API write into a vector's buffer is legal (since C++11, the buffer is guaranteed contiguous) but it does bypass the vector's size-tracking invariant, meaning the vector doesn't know the C function filled it, so `.size()` stays whatever it was. That's dangerous if someone later assumes `vec.size()` reflects valid data.
About your solution, have you considered `std::array` if the size is truly compile-time constant, or `std::unique_ptr<Type[]>` if it's runtime-determined? Either gives you exception safety for free without the manual try/catch/free/rethrow, and avoids the malloc/free C-style pattern entirely. I read the code around your `malloc`, and see the size variable are runtime-determined mostly, for example:
```
result = AudioObjectGetPropertyDataSize( id, &property, 0, NULL, &dataSize ); // line 1345
UInt32 nRanges = dataSize / sizeof( AudioValueRange ); // line 1353
```
So may be `std::unique_ptr<Type[]>` can be used. But I don't know if this `std::unique_ptr<Type[]>` is preferred in production or not, so can you please review it for me?

Thank you!

ERSUCC:
Yes, `std::unique_ptr` would do the trick. I didn't know that you could use it for `malloc`'d arrays, but I did a little test and it seemed to work as expected. I will update my solution to use `std::unique_ptr` when I get the chance. Thank you for the suggestions!

